PR #25, opened 2014-03-28
Removing line returns from branch/tag names

ritcheyer:
Encountered a bug in SourceTree where it will allow a user to copy/paste 2 lines of copy and create a branch from it. This results in the following:

```
branch with a
line break
```

Mercurial does not allow this via Command Line, so SourceTree must be doing something behind-the-scenes to convert this into an acceptable branch.

**Solution**
Removing the space from the regex and putting in it's place `\s` which will remove any whitespace, tab character or line break.

**Resolved Output**

```
Warning: sanitized branch [branch_name_on
multiple_lines] to [branch_name_on_multiple_lines]
```

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -145,7 +145,7 @@ def sanitize_name(name,what="branch"):
     return name
 
   n=name
-  p=re.compile('([[ ~^:?\\\\*]|\.\.)')
+  p=re.compile('([[~^:?\\\\\\s*]|\.\.)')
   n=p.sub('_', n)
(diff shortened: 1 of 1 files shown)


jonmills:
Just submitted a pull request for a similar problem - although it's not as elegant as your regex solution. I've no idea how the line feed got into the branch name, but it caused fast_export to crash.

frej:
This should finally have a proper solution in the [new-mapping-format branch](https://github.com/frej/fast-export/tree/new-mapping-format), check #104 for the full story.

frej:
Workaround is available in v171002, closing.

Issue #43, opened 2014-11-06
no checks in ident string


ewongbb:
I'm currently trying to convert a large mercurial repo to git using fast-export.  After running it for some time, fast-export crashes with:

fatal: Missing > in ident string: test user <test@test.com 123456778 +1200
with the following traceback:
Traceback (most recent call last):
  File "/home/cc/projs/fast-export/hg-fast-export.py", line 442, in <module>
    notes=options.notes,encoding=encoding))
  File "/home/cc/projs/fast-export/hg-fast-export.py", line 364, in hg2git
    c=export_commit(ui,repo,rev,old_marks,max,c,authors,sob,brmap,hgtags,notes,encoding)
  File "/home/cc/projs/fast-export/hg-fast-export.py", line 226, in export_commit
    export_file_contents(ctx,man,changed,hgtags,encoding)
  File "/home/cc/projs/fast-export/hg-fast-export.py", line 140, in export_file_contents
    wr(d)
  File "/home/cc/projs/fast-export/hg-fast-export.py", line 37, in wr
    wr_no_nl(msg)
  File "/home/cc/projs/fast-export/hg-fast-export.py", line 34, in wr_no_nl
    sys.stdout.write(msg)
IOError: [Errno 32] Broken pipe

c4lliope:
Picking up this long-dead thread,
I had a similar issue parsing [Mozilla's source](https://hg.mozilla.org/);
a couple `<<address@place>` cases and similar.

I am [mirroring this code base], and [in this code base] I made a couple changes;
see the [change log](https://base.assembled.app/code/hg-quick-rebase/commits/branch/main).

[mirroring this code base]: https://base.assembled.app/dependencies/hg-fast-export
[in this code base]: https://base.assembled.app/code/hg-quick-rebase

[Open an issue here] should you experience similar issues,
and link the code base you need help on.

[Open an issue here]: https://base.assembled.app/code/hg-quick-rebase/issues/new

c4lliope:
Once I searched around, I realized https://github.com/glandium/git-cinnabar handles these cases nicely,
and is a really speedy program.

I'm mirroring the code [here](https://base.assembled.app/dependencies/git-cinnabar).

PR #45, opened 2015-01-20
Crude fix - handle multiple lt or gt in author line

nickpack:
I hit the scenario where a fatal was thrown by git because of a user having multiple lt symbols in their author config, I didn't have much time to trace this through, as had a big project that I needed to convert from hg to git, so excuse the crudeness of my patch, it basically prevents fast-export from choking on:

```
* Some Dude <<some@dude.com>
* Other Dude <some@otherdude.co>>
* Another Dude <<another@dude.com>>
```

### DIFF
--- a/hg2git.py
+++ b/hg2git.py
@@ -52,6 +52,14 @@ def fixup_user(user,authors):
     # if we have 'Name <mail>' syntax, everything is fine :)
     name,mail=m.group(1),m.group(2)
 
+
+  # remove duplicated < or > - this causes fatals
+  double_ltgt = ['<<', '>>']
+
+  for dupe in double_ltgt:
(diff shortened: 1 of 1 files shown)


frej:
>    I hit the scenario where a fatal was thrown by git because of a user
>    having multiple lt symbols in their author config, I didn't have much
>    time to trace this through, as had a big project that I needed to
>    convert from hg to git, so excuse the crudeness of my patch, it
>    basically prevents fast-export from choking on:
>      * Some Dude <some@dude.com
>      * Other Dude some@otherdude.com>
>      * Another Dude <another@dude.com>

You want to use is the author map (-A) option (Look in the README).

Check issue #43 for why I won't merge this patch.

PR #46, opened 2015-06-23
Resolve empty Mercurial authors

jmichaelbako:
We ran into a crash condition where an empty Mercurial commit author could not be handled.  This resulted in the string "[devnull@localhost] [time] [tz]" being used as the input for git raw date parsing, throwing an invalid argument error.

Attempts were made to resolve the empty author through a number of authors mapping file entries, none of which worked.  Last resort was to include a blank author check in the hg2git.py and supply a default value.

### DIFF
--- a/hg2git.py
+++ b/hg2git.py
@@ -35,10 +35,12 @@ def setup_repo(url):
 
 def fixup_user(user,authors):
   user=user.strip("\"")
+  
   if authors!=None:
     # if we have an authors table, try to get mapping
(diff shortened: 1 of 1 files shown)


csdahlberg:
This change works for a repo I converted that had a few commits with empty authors. Thanks, @jmichaelbako.

frej:
I guess this can now be handled by the `-A` author mapping file, so closing this.

Issue #1, opened 2011-12-06
File "../fast-export/hg-fast-export.py", line 143, in sanitize_name


buganini:
Warning: sanitized tag [uploaded 2.0.0 for linux] to [uploaded_2.0.0_for_linux]
Exporting tag [uploaded_2.0.0_for_linux] at [hg r174] [git :175]
Traceback (most recent call last):
  File "../fast-export/hg-fast-export.py", line 384, in <module>
    options.statusfile,authors=a,sob=options.sob,force=options.force))
  File "../fast-export/hg-fast-export.py", line 325, in hg2git
    c=export_tags(ui,repo,old_marks,mapping_cache,c,authors)
  File "../fast-export/hg-fast-export.py", line 219, in export_tags
    tag=sanitize_name(tag,"tag")
  File "../fast-export/hg-fast-export.py", line 143, in sanitize_name
    if n[-1] == '/': n=n[:-1]+'_'
IndexError: string index out of range

DoDoENT:
I am getting this error when attempting to add eigen mercurial repository (ssh://hg@bitbucket.org/eigen/eigen) as git submodule using [git-hg](https://github.com/offbytwo/git-hg) which uses fast-export as submodule. The repository works correctly when used directly with hg, so I doubt it is corrupted...

frej:
> I am getting this error when attempting to add eigen mercurial
> repository (ssh://hg@bitbucket.org/eigen/eigen) as git submodule using
> git-hg which uses fast-export as submodule. The repository works
> correctly when used directly with hg, so I doubt it is corrupted...

hg clone https://bitbucket.org/eigen/eigen/

mkdir git; cd git ; git init
$ hg-fast-export.sh -r ../eigen/ -s --hgtags --hg-hash
Error: repository has at least one unnamed head: hg r5877

The repo has unnamed heads, this cannot be represented in git, see issue
#9.

PR #50, opened 2015-08-21
Handle wrong names properly

jimon:
Somehow I met repository with exactly the same issue as described here : https://github.com/frej/fast-export/issues/1
Not sure if this fix is helpful for fast-export upstream, but it wouldn't hurt to try :) If it's not helpful please just close this pull request.


frej:
This can now be handled by the `-T` tags mapping file, closing.

PR #54, opened 2015-11-03
Add filename encoding option --fe

tinysun212:
This encoding option is applied to only file names.

In my locale, default setting of Mercurial Windows version, can use two encoding for commit message and file names. Usually UTF-8 encoding for commit message and EUC-KR for file names.

This option (--fe) specifies the encoding only for file names.

### DIFF
--- a/README
+++ b/README
@@ -37,7 +37,11 @@ again.
 When a mercurial repository does not use utf-8 for encoding author
 strings and commit messages the "-e <encoding>" command line option
 can be used to force fast-export to convert incoming meta data from
-<encoding> to utf-8.
+<encoding> to utf-8. This encoding option is also applied to file names.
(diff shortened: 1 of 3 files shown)


frej:
Looks good to me, but please add a paragraph about the new option to the README-file. Also amend your commit message to explain the reason for your change, something like this:

Add filename encoding option --fe

In some locales Mercurial uses different encodings for commit messages and file names. The --fe option allows the filename encoding to be overridden.

tinysun212:
I patched README and committed new message as yours, its good.

frej:
I reformatted the commit message to 80 columns before committing in38e81367ec2960622d19509119b05983b00db2eb . Thank you for your contribution!

PR #56, opened 2015-12-12
Use README.md with GitHub markdown

mark-raymond:
The existing README file was nearly markdown, so in this pull request I have renamed it to README.md and fixed the markdown so it looks right on GitHub.


frej:
Good idea! Thanks for the contribution.

PR #59, opened 2016-02-03
Add a nicer error message if mercurial is not available

dritter:
This is from the mercurial sources itself.

It's better to get an nice error message, than a stacktrace.


frej:
That you have to have mercurial installed to run hg-fast-export should be fairly self-evident, or else you will find out when the import fails.

dritter:
:smile: True that. And I have mercurial installed (see #60 ).

frej:
Seems to be resolved (#60), closing.

PR #62, opened 2016-03-28
Add lfs support to convert hg large files into git-lfs file pointer

planestraveler:
Hello, 

Here are some simple fix that enable converting mercurial large files into git-lfs pointer. It does so by filtering .hglf files pointer. It will read the content of such file and generate a git-lfs pointer based on the [spec](https://github.com/github/git-lfs/blob/master/docs/spec.md) provided by the project. A git attributes file is also generated to take care of the lfs tracked files.

I have been able to convert mercurial repository having large files in history and be able to properly check them out.

Hope you like it.

Steve


hg-fast-export.py
```diff
@@ -149,11 +149,14 @@ def sanitize_name(name,what="branch"):
   """Sanitize input roughly according to git-check-ref-format(1)"""
 
   def dot(name):
```
frej: From the commit message it is unclear what you are trying to achieve, anyway this is not a good idea, check the newly added comment in 46bf316a3cca059d60c0701009f1fe1c6d6b51f5 for the explanation.

hg-fast-export.py
```diff
@@ -330,7 +330,7 @@ def verify_heads(ui,repo,cache,force):
   # get list of hg's branches to verify, don't take all git has
   for _,_,b in l:
     b=get_branch(b)
-    sha1=get_git_sha1(b)
+    sha1=get_git_sha1(sanitize_name(b))
```
frej: Good catch, you are credited in b7cc6ab3bfecaea47105c23739b289461c2bf347 which is basically your change, but following some refactoring.

hg-fast-export.py
```diff
@@ -122,6 +127,27 @@ def get_author(logmessage,committer,authors):
...
+    while buf:
+      hasher.update(buf)
+      size = size + len(buf)
+      buf = bfile.read(1024*1024*1024)
+  digest = hasher.hexdigest()
+  try:
+    os.makedirs("lfs/objects/%s/%s" % (digest[0:2],digest[2:4]))
+  except:
+    pass
+
```
frej: To me this looks like you look up the locally cached copy of the file and hash it. What happens if the file isn't in the user's cache? Isn't the whole point of the extension that you should not have to download all historic versions of the file? According to https://www.mercurial-scm.org/wiki/LargefilesExtension the standin in .hglf/ contains the sha1 of the file contents, why do you then have to hash the file and then copy the file into the cache on the git side?

To me a much cleaner way would be to for each standin on the hg side produce a pointer file on the git side, this should be possible to do without actually touching the large file, or?

hg-fast-export.py
```diff
@@ -122,9 +135,31 @@ def get_author(logmessage,committer,authors):
...
 
+def generate_git_lfs_pointer(filename):
+  size = 0
+  hasher = hashlib.sha256()
+  with open(filename, "rb") as bfile:
+    buf = bfile.read(1024*1024*1024)
+    while buf:
+      hasher.update(buf)
+      size = size + len(buf)
+      buf = bfile.read(1024*1024*1024)
```
frej: Why is not this a symbolic constant and do you really want a 1G buffer, would not a couple of pages be enough?

hg-fast-export.py
```diff
@@ -31,6 +31,8 @@
 # write some progress message every this many file contents written
```
frej: Please run the commit message through a spell checker and check your grammar.

frej:
I'm positive to the lfs support, but not happy with the current implementation, so address my comments and I'll merge it.

planestraveler:
Thanks for the review, I'll gladly fix and comment every modification
needed. It shall be done during the course of the next week.

hg-fast-export.py
planestraveler: There are 2 reason why I'm reading the full large file in this case. 

One is to generate the file pointer for Git-LFS, I do this by following the [Spec](https://github.com/github/git-lfs/blob/master/docs/spec.md) and we need to provide an oid and it currently only support sha256 so it is fair to assume we need to rehash the file.

Next, we need to copy the file into it's proper folder which is also sha256 based. We do this only to make sure it will be properly uploaded to the new server once the convertion done.

hg-fast-export.py
planestraveler: It is true that 1Gib of memory is a pretty big chunk, I'll make sure to reduce this and create a constant for it.

hg-fast-export.py
frej: Sorry, I missed the part where they say that git lfs uses sha256.

Still, what is your plan for when the large file is not in the hg repo's cache?

planestraveler: In this case, my intention where to output a message with step to make sure the file is downloaded in the cache and stop the process. Otherwise, I do think that making an automatic fetch may fail as you may be required to input a username and password and to my understanding the stdout is piped to git-fast-import.

But If we may know that username and password is not required to download the large files. Then it should be possible to download them automatically. Otherwise, I will stick to message and exit.

planestraveler:
Closing the pull request as #65 should replace it.

PR #65, opened 2016-04-09
Add lfs support version 2

planestraveler:
Hello again,

Following up with our discussion on #62, I've created this new pull request with requested change. This permitted to cleanup erroneous work that was done and rebase on the last master.

I've also came around a bug where the generated git attributes were not escaping space properly. I've fixed it at the same time.

Hope you like this version better. I would also point out that some improvement would also still be required as resuming the export from the last state would most likely kill the currently generated gitattribute file. My suggestion would be to pickle the current states of the file or grab it from the fast-export stream and update it without breaking the users modifications.

Hope you like this version better.

Steve


hg-fast-export.py
```diff
@@ -122,6 +127,35 @@ def get_author(logmessage,committer,authors):
       return r
   return committer
 
+lfs_pointer_buffer_size = 1024*1024
+def generate_git_lfs_pointer(filename):
+  size = 0
+  # since Mercurial storage for large files are based around the sha1 we
+  # need to rehash the file to generate the proper key for the git-lfs
+  # pointer file.
```
frej: Since Mercurial storage for large files are based around sha1 and git lfs uses sha256 we need to rehash the file to generate the oid.

hg-fast-export.py
```diff
@@ -122,6 +127,35 @@ def get_author(logmessage,committer,authors):
...
+      size = size + len(buf)
+      buf = bfile.read(lfs_pointer_buffer_size)
+  digest = hasher.hexdigest()
+  try:
+    os.makedirs("lfs/objects/%s/%s" % (digest[0:2],digest[2:4]))
+  except:
+    pass
+  # Once we find where to put the large file, we need to copy it to the
+  # cached location so the git lfs filter replace the file pointer with
+  # large file. If upload is required, it also need to be in the git
```
frej: ... lfs filter replace the file .. -> ... lfs filter replaces the file...

... it also need to be ... -> ... it also needs to be ...

hg-fast-export.py
```diff
@@ -135,6 +169,21 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding=''):
...
+      sys.stderr.write("Detected large file")
+      filename = filename[5:]
+      #should detect where the large files are located
+      lfsFileCached = lfutil.findfile(ctx.repo(), d.strip('\n'))
+      if lfsFileCached is not None:
+        d = generate_git_lfs_pointer(lfsFileCached)
+      else:
+        # Autodownloading from the mercurial repository would be an issue as there is a good chance that we may
+        # need to input some username and password. This will surely break the fast-export as there will be
+        # some unexpected output.
```
frej: ... break the fast-export as ... -> ... break fast-export as ...

hg-fast-export.py
```diff
@@ -135,6 +169,21 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding=''):
...
+        sys.stderr.write("Large file wasn't found in local cache")
+        sys.stderr.write("Please clone with --all-largefiles")
+        sys.stderr.write("Or pull with --lfrev %s" % (str(ctx.rev())))
+        sys.exit(3) # closing in the middle of import will revert everything to the last checkpoint
```
frej: Have you tested this? There are no line-feeds and no spaces between the printouts and "Or" should not be capitalized.

hg-fast-export.py
```diff
@@ -7,6 +7,11 @@
 from hg2git import setup_repo,fixup_user,get_branch,get_changeset
```
frej: Corrected commit message:

When the fast-export script finds a .hglf pointer file, it will
convert it to its counterpart git lfs pointer file. To work correctly
the mercurial large file needs to be in the user's cache, otherwise 
fast export will die after providing some instructions on how to continue.

hg-fast-export.py
```diff
@@ -16,6 +16,8 @@
 import sys
```
frej: Corrected commit message:

Generate a git attribute entry for all lfs files

For git-lfs to work properly, each file needs extra information in the
gitattributes file. The added information is used by the git filters
to replace the current pointer file by the actual large file.

For now the gitattributes file will be the same for all branches as
the current branch isn't obviously available from the current context.

frej: Corrected commit message:

Generate a git attribute entry for all lfs files

For git-lfs to work properly, each file needs extra information in the
gitattributes file. The added information is used by the git filters
to replace the current pointer file by the actual large file.

For now the gitattributes file will be the same for all branches as
the current branch isn't obviously available from the current context.

hg-fast-export.py
```diff
@@ -31,6 +33,10 @@
 # write some progress message every this many file contents written
 cfg_export_boundary=1000
 
+# those 2 dict should be pickled for future resume
+attribute_tree = {}
+attribute_file = {}
+
```
frej: Why not load the information you need from git in export_commit() and then and then pass the info down to export_file_contents() where it is updated and then finish export_commit() by updating the relevant .gitattributes files?

frej:
I have added a bunch of comments inline, the first path looks almost ready, the attribute part requires more work. Also I would like an addition to the README which describes what you have to do to use it. From the comments in the code it is clear that you just can't run fast export and expect it to work, you have to fill the cache and then do something with git lfs to publish the converted large files.

chadrik:
This looks really promising.  @planestraveler any chance that you might have time to address the review notes?  @frej do you think you might be able to address them on your end?

frej:
@chadrik: sorry I have very little time for adding features. I have updated the labels to make it clear that this will require contributor activity.

planestraveler:
I've kind of forgot about this pull request, I've added some more details about the large files in the readme section. Let me know if anything isn't still clear, it is true that the original information did assume you already know about git-lfs.

As for one comment from @frej I do agree it does seem to be a better approach. The gitattribute would be updated each time the import is resumed from the previous pass. I don't have much time to go through the refactoring right now, I would need more time to refamiliarise myself with the code.

frej:
> #65 planestraveler:add-lfs-support-v2
> #65
> #143
> #156

I'm not sure what this user is trying to do, but it's blocked now...

watsonsong:
Hello and when the lfs support can be used?

frej:
Superseded by #141.

PR #73, opened 2016-09-13
Fix PATH when using git >= 2.10

sbraz:
Since [this commit](https://github.com/git/git/commit/d323c6b6410dee0a8a2471b4dbf6d631cd3c3d4d), `git-sh-setup.sh` tries to source files in its exec path which is not
usually included in PATH.

Let me know if this is the right way to fix the issue. Without this workaround, the shell scripts fail with:
```
/usr/libexec/git-core/git-sh-setup: line 6: .: git-sh-i18n: file not found
```

### DIFF
--- a/hg-fast-export.sh
+++ b/hg-fast-export.sh
@@ -48,7 +48,8 @@ case "$1" in
       echo "$LONG_USAGE"
       exit 0
 esac
-. "$(git --exec-path)/git-sh-setup"
+git_exec_path="$(git --exec-path)"
+PATH="$PATH:$git_exec_path" . "$git_exec_path/git-sh-setup"
 cd_to_toplevel
 
 while case "$#" in 0) break ;; esac
--- a/hg-reset.sh
+++ b/hg-reset.sh
@@ -24,7 +24,8 @@ Options:
 	-r	Mercurial repository to use
 "
 
-. "$(git --exec-path)/git-sh-setup"
+git_exec_path="$(git --exec-path)"
+PATH="$PATH:$git_exec_path" . "$git_exec_path/git-sh-setup"
 cd_to_toplevel
 
 while case "$#" in 0) break ;; esac


frej:
That's a reasonable fix, but relying on the internal api of git (not your fault, fast-export has been doing this since the beginning) will probably lead to further problems in the future.

It appears that we only use git internals for discovering the repo directory and setting up the GIT_DIR environment variable. I did a quick hack to remove the sourcing of git-sh-setup.sh and do what we need (I think) by hand in  fast-export. I have pushed that as 0e7a08a716adfe1a689e3deec09010f6aaadeb2e in this [branch](https://github.com/frej/fast-export/tree/support-git-2.10), try giving it a spin and tell me if that works for you.

sbraz:
Works great, thanks, you can close this PR once it's merged into master :)
Will you also publish a new release?

frej:
Extended the commit message a bit and merged it to master. I tagged a new release as there is at least one user it works for :)

klemens:
The "quick hack" does not work for bare git repos, as `cd $(git rev-parse --show-toplevel)` changes to the home directory using the default dash on a current debian, because `git rev-parse --show-toplevel` is empty.

I suggest adding quotes to it, as `cd ""` just keeps the current directory here. This should also help with paths that contain spaces. 😉 However, while is seems to work in bash, dash and zsh, I am not sure if it works for every shell (it fails eg for tcsh, although you probably don't want to use that as `/bin/sh`…).

```diff
diff --git a/hg-fast-export.sh b/hg-fast-export.sh
index f3f58f6..9f9b1a1 100755
--- a/hg-fast-export.sh
+++ b/hg-fast-export.sh
@@ -49,7 +49,7 @@ case "$1" in
       exit 0
 esac
 
-cd $(git rev-parse --show-toplevel) \
+cd "$(git rev-parse --show-toplevel)" \
     || (echo "Could not find git repo" ; exit 1)
 GIT_DIR=$(git rev-parse --git-dir) || exit 1
 
diff --git a/hg-reset.sh b/hg-reset.sh
index d7fbff7..f09e130 100755
--- a/hg-reset.sh
+++ b/hg-reset.sh
@@ -24,7 +24,7 @@ Options:
        -r      Mercurial repository to use
 "
 
-cd $(git rev-parse --show-toplevel) \
+cd "$(git rev-parse --show-toplevel)" \
     || (echo "Could not find git repo" ; exit 1)
 GIT_DIR=$(git rev-parse --git-dir) || exit 1
 
```
Note: Assigning `GIT_DIR` also works for paths with spaces even without quotes.

PR #76, opened 2016-09-14
Prepend the commit message with the hg branch name

cdimech:
Add an option to prepend the commit message with the hg branch name by specifying the switch -c.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -179,7 +179,7 @@ def strip_leading_slash(filename):
   return filename
 
 def export_commit(ui,repo,revision,old_marks,max,count,authors,
-                  branchesmap,sob,brmap,hgtags,encoding='',fn_encoding=''):
+                  branchesmap,sob,brmap,hgtags,encoding='',fn_encoding='',commitmsg=False):
   def get_branchname(name):
     if brmap.has_key(name):
       return brmap[name]
@@ -201,6 +201,11 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
   if sob:
     wr('author %s %d %s' % (get_author(desc,user,authors),time,timezone))
   wr('committer %s %d %s' % (user,time,timezone))
+
+  #check if the desc does not start with the branch name before modifying the commit msg.
+  if commitmsg and not desc.startswith(branch):
+    desc=branch + ': ' + desc
+
   wr('data %d' % (len(desc)+1)) # wtf?
   wr(desc)
   wr()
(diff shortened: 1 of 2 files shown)


hg-fast-export.py
```diff
@@ -359,7 +364,7 @@ def verify_heads(ui,repo,cache,force,branchesmap):
 
 def hg2git(repourl,m,marksfile,mappingfile,headsfile,tipfile,
            authors={},branchesmap={},tagsmap={},
-           sob=False,force=False,hgtags=False,notes=False,encoding='',fn_encoding=''):
+           sob=False,force=False,hgtags=False,notes=False,encoding='',fn_encoding='',commitmsg=False):
```
frej: The name of the parameter does not match its purpose, from the name I could never guess that it is controlling the addition of a prefix to the title line of the commit message.

hg-fast-export.py
```diff
@@ -447,6 +452,8 @@ def bail(parser,opt):
       help="Read branch map from BRANCHESFILE")
   parser.add_option("-T","--tags",dest="tagsfile",
       help="Read tags map from TAGSFILE")
+  parser.add_option("-c",action="store_true",dest="commitmsg",
```
frej: Is '-c' the right flag, likewise, the name 'commitmsg' does not match the option.

frej:
I don't see the use for this functionality. If you do not prefix commit titles with the branch name on the hg side, why do you need it on the git side? Branches are converted so you do not lose this information during conversion.

The purpose of fast-export is to convert from hg to git. Fast-export supports mangling of author and branch names to allow for a successful conversion in cases where hg and git are incompatible. Rewriting commit messages (outside of an encoding change) is outside the scope of fast-export, if you want to do arbitrary rewriting of commit messages, use git-filter-branch following fast-export.

Apart from the philosophical objection I have added some in-line comments in patch, but don't spend any time addressing them unless you convince me on the need for the proposed functionality.

cdimech:
Thanks for your response.

The reason why this new feature is implemented is to keep track of what the original branch name was with respect to a changeset. According to this article (http://stevelosh.com/blog/2009/08/a-guide-to-branching-in-mercurial/#comparison-to-git_2), branch information is never stored as part of Git changeset metadata. Git's commit object contains a pointer to commit/commits which directly comes/came before the given commit, making it more difficult to infer the branch name from an independent commit. The branch name is crucial for us since it refers to a JIRA task thus enabling us to trace the history of a particular feature using the branch name as part of the changeset metadata. This information would be lost if we do not add such prefix while migrating from Hg to Git.

frej:
I see, the motivation for this proposed change is that your bug-tracker uses the hg branch name to track issues (this sounds almost like Gerrit's `Change-Id: ` field)? I still think the proposed change is too specific and not very "gitish". Metadata, in the git world, is normally recorded as `<Field>: <value>` pairs in the commit message.

Doing it _right_ and not only useful for this particular case would require functionality to transform commit messages to add customized meta-data. There is already a tool to do that :  `git filter-branch`.

If `git branch --contains <commit>` is not sufficient to retrieve the information you need in this case, I would suggest that you use fast-export's `--hg-hash`-flag combined with a `git filter-branch` pass. The filter can then retrieve the note containing the hg hash and then look up the branch name and add it to the commit message.

frej:
Closing this issue as I don't see the need for this functionality. The wanted end result can be achieved greater flexibility with already existing functionality and tools.

PR #78, opened 2016-09-21
make annotated tags, not weak tags

imz:
It seems that hg has some meta info about a tag anyway: the tagger and date, which is different from the previous commit that is being tagged. It could be saved correctly in git in an annotated tag object.

Nevertheless, the message is empty according to this patch.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -287,8 +287,11 @@ def export_tags(ui,repo,old_marks,mapping_cache,count,authors,tagsmap):
           ' %s at r%d\n' % (tag,rev))
       continue
     sys.stderr.write('Exporting tag [%s] at [hg r%d] [git %s]\n' % (tag,rev,ref))
-    wr('reset refs/tags/%s' % tag)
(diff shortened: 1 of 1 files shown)


imz:
Please check whether my understanding is correct before merging it.

The idea may be correct, but the meta info could be wrongly taken from a wrong hg commit (not the tag)...

frej:
I like it, I'll take a look.

imz:
I have looked a bit on the results, and have even run some special tests (hg commit, then hg tag wih a different date), and this code seems not to give what I supposed. The date is taken not from the tagging, but the previous one, i.e. from the tagged commit.

This code needs to be improved, although I like the idea.

I haven't yet come up with a code with the wanted behavior, because I had no time for this.

frej:
Keep digging, as I said, I think this will be a good addition. If you make it work I'll accept the patch. One thing though, don't make this the mandatory behaviour. If users have already existing "plain" tags in their converted repos, other users must be able to redo the conversion on their end and end up with the same result.

Issue #91, opened 2017-05-12
Crash when hg branch name is /?


ermshiperete:
When hg-fast-export encounters the branch name '/?' it crashes.

hg-fast-export deals with some branch names that are invalid in git:

    Hg branch  -> git branch 
    a? -> a_
    a/ -> a_
    a/b -> a/b
    a/? -> a/_
    ?a -> _a
    a. -> a_
    a.b -> a_b
    .a -> _a
    / -> _

hg-fast-export crashes when the mercurial repo contains branch names starting with a `/` (unless the name is '/'):

    /a

Even providing a branch mapping file doesn't help.

ermshiperete:
Stacktrace with latest master branch:

	Loaded 146 authors
	Traceback (most recent call last):
	  File "$HOME/bin/fast-export/hg-fast-export.py", line 504, in <module>
	    notes=options.notes,encoding=encoding,fn_encoding=fn_encoding))
	  File "$HOME/bin/fast-export/hg-fast-export.py", line 382, in hg2git
	    if not verify_heads(ui,repo,heads_cache,force,branchesmap):
	  File "$HOME/bin/fast-export/hg-fast-export.py", line 340, in verify_heads
	    sanitized_name=sanitize_name(b,"branch",branchesmap)
	  File "$HOME/bin/fast-export/hg-fast-export.py", line 168, in sanitize_name
	    n='/'.join(map(dot,n.split('/')))
	  File "$HOME/bin/fast-export/hg-fast-export.py", line 161, in dot
	    if name[0] == '.': return '_'+name[1:]
	IndexError: string index out of range

frej:
I created a dummy hg repo containing branches named as in your initial report. Running with a branch mapping file (-B) containing:

```text
a?=valid-0
a/=valid-1
a/b=valid-2
a/?=valid-3
?a=valid-4
a.=valid-5
a.b=valid-6
.a=valid-7
/=valid-8
```
works fine for me. Please provide a step-by step instruction for creating your troublesome hg repo, the contents of your mapping file, and the exact command line you use to run fast-export.

ermshiperete:
As you wish:

	~/$ cd ~/bin/fast-export/
	~/bin/fast-export[remotes/origin/HEAD]/$ git log -1
	commit fb05ce5b7bb549ef5dbfeaa4ec5740b78c4adcbe
	Author: Emilian Bold <emilian.bold@gmail.com>
	Date:   Fri Dec 30 21:47:01 2016 +0200

[...]

PR #92, opened 2017-05-12
Fix bug when hg branch name is /? (#91)

ermshiperete:
Previously hg-fast-export crashed when it encountered the branch name
/?. This change fixes it.

---
This change is [Reviewable](https://reviewable.io/reviews/frej/fast-export/92)

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -158,12 +158,20 @@ def sanitize_name(name,what="branch", mapping={}):
   # work to do manually, write a tool that does it for you.
 
   def dot(name):
-    if name[0] == '.': return '_'+name[1:]
+    if name[0] == '.':
+      if len(name) > 1:
+        return '_'+name[1:]
+      else: return '_'
     return name
 
   n=mapping.get(name,name)
   p=re.compile('([[ ~^:?\\\\*]|\.\.)')
   n=p.sub('_', n)
+  if n[0] in ('/', '.'):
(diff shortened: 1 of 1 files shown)


frej:
It's not in fast-exports interests to accumulate arbitrary transforms to try to fix up branch-names not allowed by git (I'm pretty sure your example is handled by fast-export, bit git fast-import aborts on it).  Unfortunately we are stuck with some built-in remappings to maintain compatibility with incremental conversions that might be out in the world.  Have a look at the "Tag and Branch Naming" section of the readme to see how you can rename branch names not allowed by git without having to modify fast-export.

ermshiperete:
Thanks for the hint to the "Tag and Branch Naming" section.

I agree that it's less than ideal to put in special code for dealing with invalid branch names. Would be nice if fast-export wouldn't crash though... And even better if it would output a hint to the "Tag and Branch Naming" section instead.

Re: my example handled by fast-export: I got a crash on exactly the lines I fixed...

ermshiperete:
I added the callstack to #91

ermshiperete:
@frej Unfortunately providing a branch naming file doesn't help when the branch name starts with '/' (unless the name of the hg branch is just '/'). If it starts with a slash it'll always crash.

ermshiperete:
Better fix provided in 4bb50bb3fba1b86decf62d1f8763d56e34ac902c

Issue #93, opened 2017-05-15
hg-fast-export.sh fails when symlinked


cumuru:
As I have to migrate hg repos from time to time I‘d like to have `hg-fast-export` in my path. For that purpose I cloned the repository in `/opt` and created a symlink `ln -s /opt/fast-export/hg-fast-export.sh /usr/local/bin/hg-fast-export`. Though I doubt it will matter: My OS is Ubuntu xenial.

Unfortunately this results in `hg-fast-export.py` being expected in `/usr/local/bin`. I added a `realpath` to determination of `$ROOT` in `hg-fast-export.sh` which does work. If there‘s a chance to get it merged I could create a PR. Are there any downsides on using `realpath`? Should `readlink` be preferred?

frej:
I'd welcome a patch that makes fast-exporth do this correctly. I think readlink should be preferred as I recall a problem from work where realpath was not available on a colleague's Mac.

cumuru:
Hum, no Mac over here, I can‘t check that. Nonetheless still willing to contribute.
Would something like
```
ROOT="$(dirname "$(greadlink -f "$(which "$0")" 2> /dev/null || readlink -f "$(which "$0")")")"
```
do the trick? It works on my Linux machine at least.
Or is there a more sophisticated way to do command fallback?

PR #94, opened 2017-05-20
Make hg-fst-export.sh callable via a symbolic link

cumuru:
Calling hg-fast-export.sh via symlink used to fail because
the script would look for hg-fast-export.py in symlink‘s
folder. This patch adds symlink resolution using readlink and
greadlink with fallback to support MacOs. That way you can
safely add a symlink to hg-fast-export.sh somewhere in you PATH.

Fixes https://github.com/frej/fast-export/issues/93

### DIFF
--- a/hg-fast-export.sh
+++ b/hg-fast-export.sh
@@ -3,7 +3,11 @@
 # Copyright (c) 2007, 2008 Rocco Rutte <pdmef@gmx.net> and others.
 # License: MIT <http://www.opensource.org/licenses/mit-license.php>
 
-ROOT="$(dirname "$(which "$0")")"
+READLINK="readlink"
+if command -v greadlink > /dev/null; then
+  READLINK="greadlink" # Prefer greadlink over readlink
+fi
+ROOT="$(dirname "$($READLINK -f "$(which "$0")")")"
 REPO=""
 PFX="hg2git"
 SFX_MAPPING="mapping"


hg-fast-export.sh
```diff
@@ -3,7 +3,7 @@
 # Copyright (c) 2007, 2008 Rocco Rutte <pdmef@gmx.net> and others.
 # License: MIT <http://www.opensource.org/licenses/mit-license.php>
 
-ROOT="$(dirname "$(which "$0")")"
+ROOT="$(dirname "$(greadlink -f "$(which "$0")" 2> /dev/null || readlink -f "$(which "$0")")")"
```
frej: You could do this in two stages to make it a bit simpler to understand what's happening and why:

```shell
READLINK="readlink"
if command -v greadlink > /dev/null; then
  READLINK="greadlink" # Prefer greadlink over readlink
fi
ROOT="$(dirname "$($READLINK -f "$(which "$0")")")"
```

cumuru: Sorry, I‘m no native speaker :-). I like your suggestion using an additional variable for its readability - I‘m going to alter my code accordingly.

cumuru:
I changed commit message as suggested and squashed the two commits. Can it be merged now?

frej:
You forgot to add the missing definite and indefinite articles in the commit message body. Instead of doing one more round of update and review I'll just amend it before pushing, I hope you won't mind.

cumuru:
I don‘t mind at all. I didn‘t notice it took the wrong commit message when squashing the two commits. Thanks for merging!

PR #95, opened 2017-05-28
Filter file contents, Was: Suggestion: add --as-binary option to convert line endings

atykhyy:
Converting line endings in text files to LF is often needed when migrating from Mercurial to Git. Right now it is usually done by git filter-branch and other heavy-weight approaches, but it is very easy to do in fast-export provided that binary files have distinct filenames (e.g. distinct extensions). This change adds an option `--as-binary` that turns on auto-conversion of line endings to LF and takes a list of binary file extensions, e.g. `hg-fast-export ..\repo --as-binary=.so,.dll`. As an implementation side-effect, dot-files (e.g. `.hgignore`) are treated as text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,27 @@ name the -B and -T options allow a mapping file to be specified to
 rename branches and tags (respectively). The syntax of the mapping
 file is the same as for the author mapping.
 
+Content filtering
+-----------------
+
+hg-fast-export supports filtering the content of exported files.
+The filter is supplied to the --filter-contents option. hg-fast-export
+runs the filter for each exported file, pipes its content to the filter's
+standard input, and uses the filter's standard output in place
+of the file's original content. The prototypical use of this feature
+is to convert line endings in text files from CRLF to git's preferred LF:
+
+```
+-- Start of crlf-filter.sh --
+#!/bin/sh
+# $1 = pathname of exported file relative to the root of the repo
+# $2 = Mercurial's hash of the file
+# $3 = "1" if Mercurial reports the file as binary, otherwise "0"
+
+if [ "$3" == "1" ]; then cat; else dos2unix; fi
+-- End of crlf-filter.sh --
+```
+
 Notes/Limitations
 -----------------
 
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -123,7 +123,7 @@ def get_author(logmessage,committer,authors):
       return r
   return committer
 
-def export_file_contents(ctx,manifest,files,hgtags,encoding=''):
+def export_file_contents(ctx,manifest,files,hgtags,encoding='',filter_contents=None):
   count=0
   max=len(files)
   for file in files:
@@ -131,11 +131,24 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding=''):
     if not hgtags and file == ".hgtags":
       sys.stderr.write('Skip %s\n' % (file))
       continue
-    d=ctx.filectx(file).data()
     if encoding:
       filename=file.decode(encoding).encode('utf8')
     else:
       filename=file
+    file_ctx=ctx.filectx(file)
+    d=file_ctx.data()
+    if filter_contents:
+      import subprocess
+      filter_cmd=filter_contents + [filename,node.hex(file_ctx.filenode()),'1' if file_ctx.isbinary() else '0']
+      try:
+        filter_proc=subprocess.Popen(filter_cmd,stdin=subprocess.PIPE,stdout=subprocess.PIPE)
+        d,_=filter_proc.communicate(d)
+      except:
+        sys.stderr.write('Running filter-contents %s:\n' % filter_cmd)
+        raise
+      filter_ret=filter_proc.poll()
+      if filter_ret:
+        raise subprocess.CalledProcessError(filter_ret,filter_cmd)
     wr('M %s inline %s' % (gitmode(manifest.flags(file)),
                            strip_leading_slash(filename)))
     wr('data %d' % len(d)) # had some trouble with size()
@@ -185,7 +198,7 @@ def strip_leading_slash(filename):
   return filename
 
 def export_commit(ui,repo,revision,old_marks,max,count,authors,
-                  branchesmap,sob,brmap,hgtags,encoding='',fn_encoding=''):
+                  branchesmap,sob,brmap,hgtags,encoding='',fn_encoding='',filter_contents=None):
   def get_branchname(name):
     if brmap.has_key(name):
       return brmap[name]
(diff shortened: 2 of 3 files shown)


frej:
My spontaneous reaction is that this is outside the scope of fast-export. Filter branch is the tool to use when you have to mangle the content of the version controlled files.

Additionally I see a number of problems with the implementation (just to be clear, even if these problems are addressed I'm not very likely to include the proposed functionality):

* The name of the new flag does not in any way indicate that it does line ending conversion for all files not listed by the flag.

* The only  supported conversion is from CRLF to LF, what about the reciprocal? What about CR<-->LF if your source files are from a classic Mac? Which conversions should be supported?

* How do you handle binary files without an extension? Shouldn't the pattern be a proper glob?

atykhyy:
Indeed, using filter-branch after fast-import is the usual approach. However, converting line endings is the most common, and often the only, post-import operation when migrating repos from hg to git. Wouldn't it be convenient to take care of that with just an extra flag or two and not have to filter-branch (which is quite slow) afterwards?

I believe restricting conversion to target just LF is justified because git uses LF line endings for text files internally, and it takes some contortions to force git to store them in a different format even if one wanted it to. I implemented the rest of your very reasonable suggestions.

frej:
As I said, I think this is outside the scope of fast-export. Since your first proposal this simple filter has already been extended to handle one more line ending. Pretty soon someone else will want support for converting character encodings, running clang-format on c-code, stripping old forgotten $Id$-tags, etc, and I don't want fast-export to evolve into emacs or acquire more flags than ls :)

What about adding a new flag `--filter-contents <filter>` where `<filter>` is an executable filter which is called as `<filter> <file-path>`? That way the filter can do whatever it wants. I would definately merge such a patch as it adds a lot of flexibility for the small increase in fast-export complexity.

Himura2la:
`--filter-contents` is cool. Especially if we create an example with line endings conversion using this feature.

atykhyy:
How about this one? I added a `<binary>` flag to the filter's command line arguments since Mercurial `filectx` provides it.

Himura2la:
@atykhyy can you please add this option to `.sh`, I'm not sure it's safe to use the `.py` directly, there seems to be many checks and preparations in `.sh`

Himura2la:
OK, thanks for quick reaction. But I think I don't understand how to use the filter. I guess the filtering script has no access to the filteres file, it simply executes in a git repo with empty index... So HOW to properly change the contens? Should I amend each commit?

atykhyy:
Yes, amend each file. There is no way the filtering script can have direct access to files in Git, they aren't imported until the output of `hg-fast-export` gets to `git fast-import`. I suppose the script can access files in Mercurial using the provided path and hash value, but this should not be necessary. If your goal is just to convert line endings, and you have no binary files in your repo, `--filter-contents=dos2unix` should work; if `dos2unix` is confused about its arguments, and/or if you have binary files, use a simple script like
```
#!/bin/sh
if "$3"; then dos2unix; else cat; fi
```
instead. Send the output of `hg-fast-export` to a file to check the results. I have used intermediate scripts that parse the output of `hg-fast-export` and pipe it on to `git fast-import` to split Mercurial repositories, fixing paths in `.csproj` files etc. In fact I'd like to add `--filter-names` to enable the filter script(s) to skip and rename files, functionality that is not easy to squeeze into `--filter-contents` as it accepts the (possibly binary) exported file as its standard input and writes the filtered file to its standard output.

Himura2la:
Ah, so the file contents are piped into the script and the stdout is used as file contents! That's super convenient, I came up with the following script for documentation:
```sh
#!/bin/sh
#
# This script is executed in your **git** repository root
# for every file in every commit.
#
# The file contents is piped into it, and its output is used
# as file contents. It also takes three arguments:

path=$1
hash=$2
is_binary=$3

# Let's perform a line endings conversion without touching the binaries:

if [ "$is_binary" -eq "1" ]; then cat; else
    dos2unix
fi
```

hg-fast-export.py
```diff
@@ -123,19 +126,31 @@ def get_author(logmessage,committer,authors):
...
       continue
-    d=ctx.filectx(file).data()
     if encoding:
       filename=file.decode(encoding).encode('utf8')
     else:
       filename=file
+    f=ctx.filectx(file)
+    d=f.data()
+    if filter_contents:
+      a=filter_contents + [filename,binascii.hexlify(f.filenode()),'1' if f.isbinary() else '0']
```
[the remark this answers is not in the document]
atykhyy: I'll look for one.

atykhyy:
OK. What do you think about `--filter-names` or similar to rename/exclude files ([see above](https://github.com/frej/fast-export/pull/95#issuecomment-396533998))?

Himura2la:
If the `--filter-names` would be used for this https://github.com/frej/fast-export/issues/129, I'd be happy

frej:
@atykhyy: I think `--filter-names` is a good idea, I have created #130 for it, in order to not mix up two mostly orthogonal features.

atykhyy:
Added example to readme and squashed.

README.md
```diff
@@ -99,6 +99,27 @@ name the -B and -T options allow a mapping file to be specified to
...
+is to convert line endings in text files from CRLF to git's preferred LF:
+
+```
+-- Start of crlf-filter.sh --
+#!/bin/sh
+# $1 = pathname of exported file relative to the root of the repo
+# $2 = Mercurial's hash of the file
+# $3 = "1" if Mercurial reports the file as binary, otherwise "0"
+
+if "$3"; then dos2unix; else cat; fi
```
frej: Try
```
X=1
if "$X" ; then echo true ; else echo false ; fi
```
and you'll see that this is not valid sh, you need to do `if [ "$3" == "1" ]; ...`

hg-fast-export.py
```diff
@@ -123,19 +123,32 @@ def get_author(logmessage,committer,authors):
...
-    d=ctx.filectx(file).data()
     if encoding:
       filename=file.decode(encoding).encode('utf8')
     else:
       filename=file
+    f=ctx.filectx(file)
+    d=f.data()
+    if filter_contents:
+      import subprocess
+      a=filter_contents + [filename,node.hex(f.filenode()),'1' if f.isbinary() else '0']
```
frej: `a`? give it a more descriptive name please, what about `filter_cmd`?

hg-fast-export.py
```diff
@@ -123,19 +123,32 @@ def get_author(logmessage,committer,authors):
...
     # Skip .hgtags files. They only get us in trouble.
     if not hgtags and file == ".hgtags":
       sys.stderr.write('Skip %s\n' % (file))
       continue
-    d=ctx.filectx(file).data()
     if encoding:
       filename=file.decode(encoding).encode('utf8')
     else:
       filename=file
+    f=ctx.filectx(file)
```
frej: `f` what is it? Call it `file_info` so I that I don't mix it up with all the other file related variables.

hg-fast-export.py
```diff
@@ -123,19 +123,32 @@ def get_author(logmessage,committer,authors):
...
+    d=f.data()
+    if filter_contents:
+      import subprocess
+      a=filter_contents + [filename,node.hex(f.filenode()),'1' if f.isbinary() else '0']
+      try:
+        p=subprocess.Popen(a,stdin=subprocess.PIPE,stdout=subprocess.PIPE)
```
frej: `subproc` instead of `p`

atykhyy: Done.

atykhyy:
Fixed variable naming.

frej:

PR #107, opened 2018-01-24
Quote $PYTHON variable to support paths with spaces

mfrnd:


frej:

PR #108, opened 2018-01-30
sanitize hg branch /name => root/name

Alex-Aralis:
Git rejects branch names with leading slash.  This pr converts `/branch_name` to `root/branch_name`.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -170,6 +170,9 @@ def sanitize_name(name,what="branch", mapping={}):
   p=re.compile('_+')
   n=p.sub('_', n)
 
+  if n[0] == '/':
+    n='root'+n
+  
   if n!=name:
     sys.stderr.write('Warning: sanitized %s [%s] to [%s]\n' % (what,name,n))
   return n


frej:
> Git rejects branch names with leading slash. This pr converts
> /branch_name to root/branch_name.

Since I took over the maintenance of fast-export I try to not inflict
arbitrary naming policy on the user, instead I try to provide mechanisms
allowing the user to apply their own policy. Your patch implements just
such an arbitrary policy and it can equally well be implemented with a
branch mapping file (command line option -B). As the comment starting on
line 151 (soon updated to mention this case) hints: some shell scripting
should let you generate a mapping file.

Alex-Aralis:
@frej Sure but the whole point of this project is to convert Mercurial to Git.  Why would the default sanitization method allow the creation of non git conformant branch names? 

If a user desires a different transformation than the default, they can simply use a branch mapping file to eliminate leading slashes and the leading slash sanitization will never be "inflicted upon" them.

frej:
> Sure but the whole point of this project is to convert Mercurial to
> Git. Why would the default sanitization method allow the creation of
> non git conformant branch names?

The current default does not allow the creation of non git conformant
branch names, it makes git fast-import crash, and that is a feature. The
user is then forced to deal with the impedance mismatch between git and
hg.

> If a user desires a different transformation than the default, they
> can simply use a branch mapping file to eliminate leading slashes and
> the leading slash sanitization will never be "inflicted upon" them.

The problem with this patch is that while solving your problem, it
introduces another, what happens if the user happens to have a branch
named "root"?

The only safe way to avoid this kind of problem is adding a prefix to
_all_ branch names (or some other way to eliminate clashes) and that
can't be done in a backwards compatible way. In my opinion, providing a
mechanism and not a policy to deal with these kinds of problems is the
right way. Besides, it avoids bikeshedding about what the policy should
be.

Alex-Aralis:
Currently all sorts of assumptions are already being made in that sensitization method.  Spaces and dots are converted in a way that could very well create similar name collision to what you describe.  For instance if branches `no space` and `no_space` both exist in hg then the sanitizer will cause a collision. 

If name collision are a concern they can be resolved in a more general way via a branch name deduplication step.

-----------

I don't agree that this change would affect backwards compatibility.  There are a few cases:

### The user has no branches that begin with leading slashes

In this case obviously behavior remains the same.

### The user has branches that begin with leading slashes.

This user is either using the "crashing feature" or the branch mapping feature.

#### Crashing

Either the leading slash sanitation now fixes their problem or causes another crash, this time because of a branch name collision.  Either the user was crashing before and is still crashing, or was crashing and now exporting successfully.  Neither case is a backwards compatibility concern.

#### Branch mapping

The branch mapping is applied before the leading slash fix, so if the branch mapping fixes the issue the `root` will never be prepended and behavior remains the same.

-------------------

This covers all the cases.  It seems to me that in no case is backwards compatibility disturbed.

frej:
> Currently all sorts of assumptions are already being made in that
> sensitization method. Spaces and dots are converted in a way that
> could very well create similar name collision to what you
> describe. For instance if branches no space and no_space both exist in
> hg then the sanitizer will cause a collision.

True, but the default sanitation has been there since the beginning and
I don't want to keep making it more and more "intelligent" as by doing
that I'm adding policy. It's better to let the user know that they are
using branch names that are not gittish and let them decide what to do
with them.

> I don't agree that this change would affect backwards
> compatibility. There are a few cases:

I agree in this case, but the more complex the sanitation becomes, the
harder it becomes to know. Can you, on the top of your head, tell me if
the following scenario is sure not to trash your converted repo?

Assume that upstream HG has a branch called Root/foo (that has not been
updated in ages) and is running an incremental import on Mac. Then
fast-export is upgraded. Upstream adds a branch "/foo", an incremental
import is run. Will the old Root/foo be updated to point to "/foo"'s
head or will something else happen?

Alex-Aralis:
It seems like the real problem is a lack of duplication checking in `brmap`. 

Making sure that no two values of `brmap` are identical on addition to the map would eliminate that entire class of error.

https://github.com/frej/fast-export/blob/master/hg-fast-export.py#L192

frej:
Provided it is done in a way that produces the same result for both incremental and full imports. I agree that it would be nice to have as fast-export could warn before risking trashing the repo, patches wellcome.

PR #109, opened 2018-02-13
Update usage section example commands

ceqi-zz:
Change <repo> to <local-repo> so that it's clear that we invoke from a local repository;
Add 'git checkout HEAD' command as we need to run it as the final step.

Thanks


frej:
I like it, thank you for your contribution.

mjh10brooke-max:
8762fee403c8cbfa621e49974f45b1b87cb30d0a

mjh10brooke-max:
Track and trace

PR #110, opened 2018-02-16
Replace \ and / with _ in branch names to avoid errors when exporting…

okobets:
… sub branches when parent branch exists.

Solves issue when you have branch named x and sub branches named x/y.
Before, it would create a file for x and when encountered sub branch under x would try to create x as directory, obviously failing because it already exists as file.

This replaces both type of slashes to underscore and was the only thing I could do to export a huge repo with lots of badly named branches.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -196,6 +196,12 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
 
   branch=get_branchname(branch)
 
+  if "/" in branch:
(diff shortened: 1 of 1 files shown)


frej:
Sanitizing should happen (if it should happen at all) in sanitize_name() as called by get_branchname(). Please read the comment in sanitize_name() and the discussion in issue #108 for why this patch will not be merged.

I would accept a patch that extends the mapping file syntax to allow regexp-based rewriting...

PR #131, opened 2018-06-17
Backward compatibility with Mercurial < 4.6

Himura2la:
It needs testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ first time.
 System Requirements
 -------------------
 
-This project depends on Python 2.7 and the Mercurial 4.6 package. If
+This project depends on Python 2.7 and the `mercurial` package. If
 Python is not installed, install it before proceeding. The Mercurial
 package can be installed with `pip install mercurial`.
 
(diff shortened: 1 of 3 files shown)


Himura2la:
I know it's a very bad architecture, but I hope such a breaking changes in the API happens not too often. There's a good chance that the good abstraction layer is an overkill. I think I still can improve the beauty of the `repo.lookup` solution and get rid of the version flag

frej:
Just to make it clear and so you don't spend time on doing work that is
never going to get merged, I'm not going to merge backwards
compatibility code that complicates the logic of the main code-base.

If you don't need the new features, just stay on the old release. If you
need the new features but can't meet the requirements or can't wait
until your distribution upgrades, backport them to the most recent
release.

Himura2la:
It's no problem for me, but I expect that many people will be unable to use this tool and don't know why nothing works

Himura2la:
I still believe the transfer should be smooth.

frej:
The developer does not have the time to test and maintain fast-export for more than the most recent Mercurial version.

PR #144, opened 2018-09-11
Plugin System

jhenkens:
Throwing these out there for discussion, before I do any work to potentially isolate them if desired.

After seeing the changes to run a shell script, I tried doing it on a repo with a few hundred thousand commits, to run dos2unix on the files, and it resulted in a 5-10x slowdown, which was unnacceptable. To counter this, I perform a simple string replacement on non-binary files via python, with nearly zero slowdown.

Second, mercurial stores information historically in the branch name. For some, this information would be bad to lose, so I added a flag to either prepend or append the branch name to the commit message during the conversion.

Let me know what you think of these features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,54 @@ if [ "$3" == "1" ]; then cat; else dos2unix; fi
 -- End of crlf-filter.sh --
 ```
 
+
+Plugins
+-----------------
+
+hg-fast-export supports plugins to manipulate the file data and commit
+metadata. The plugins are enabled with the --plugin option. The value
+of said option is a plugin name (by folder in the plugins directory),
+and optionally, and equals-sign followed by an initialization string.
+
+There is a readme accompanying each of the bundled plugins, with a
+description of the usage. To create a new plugin, one must simply
+add a new folder under the `plugins` directory, with the name of the
+new plugin. Inside, there must be an `__init__.py` file, which contains
+at a minimum:
+
+```
+def build_filter(args):
+    return Filter(args)
+
+class Filter:
+    def __init__(self, args):
+        pass
+        #Or don't pass, if you want to do some init code here
+```
+
+
+```
+commit_data = {'branch': branch, 'parents': parents, 'author': author, 'desc': desc}
+
+def commit_message_filter(self,commit_data):
+```
+The `commit_message_filter` method is called for each commit, after parsing
+from hg, but before outputting to git. The dictionary `commit_data` contains the
+above attributes about the commit, and can be modified by any filter. The
+values in the dictionary after filters have been run are used to create the git
+commit.
+
+```
+file_data = {'filename':filename,'file_ctx':file_ctx,'d':d}
+
+def file_data_filter(self,file_data):
+```
+The `file_data_filter` method is called for each file within each commit.
+The dictionary `file_data` contains the above attributes about the file, and
+can be modified by any filter. `file_ctx` is the filecontext from the
+mercurial python library.  After all filters have been run, the values
+are used to add the file to the git commit.
+
 Notes/Limitations
 -----------------
 
--- a/README.md
+++ b/README.md
@@ -145,6 +145,8 @@ class Filter:
         #Or don't pass, if you want to do some init code here
 ```
 
+Beyond the boilerplate initialization, you can see the one of the
+defined filter methods in the [dos2unix](./plugins/dos2unix) plugin.
 
 ```
 commit_data = {'branch': branch, 'parents': parents, 'author': author, 'desc': desc}
--- a/README.md
+++ b/README.md
@@ -145,8 +145,9 @@ class Filter:
         #Or don't pass, if you want to do some init code here
 ```
 
-Beyond the boilerplate initialization, you can see the one of the
-defined filter methods in the [dos2unix](./plugins/dos2unix) plugin.
+Beyond the boilerplate initialization, you can see the two different
+defined filter methods in the [dos2unix](./plugins/dos2unix) and
+[branch_name_in_commit](./plugins/branch_name_in_commit) plugins.
 
 ```
 commit_data = {'branch': branch, 'parents': parents, 'author': author, 'desc': desc}
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -11,6 +11,7 @@ from optparse import OptionParser
 import re
 import sys
 import os
+import pluginloader
 
 if sys.platform == "win32":
   # On Windows, sys.stdout is initially opened in text mode, which means that
@@ -123,7 +124,7 @@ def get_author(logmessage,committer,authors):
       return r
   return committer
 
-def export_file_contents(ctx,manifest,files,hgtags,encoding='',filter_contents=None):
+def export_file_contents(ctx,manifest,files,hgtags,encoding='',filter_contents=None,plugins={}):
   count=0
   max=len(files)
   for file in files:
@@ -149,6 +150,15 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding='',filter_contents=N
       filter_ret=filter_proc.poll()
       if filter_ret:
         raise subprocess.CalledProcessError(filter_ret,filter_cmd)
+
+    if plugins and plugins['file_data_filters']:
+      file_data = {'filename':filename,'file_ctx':file_ctx,'data':d}
+      for filter in plugins['file_data_filters']:
+        filter(file_data)
+      d=file_data['data']
+      filename=file_data['filename']
+      file_ctx=file_data['file_ctx']
+
     wr('M %s inline %s' % (gitmode(manifest.flags(file)),
                            strip_leading_slash(filename)))
     wr('data %d' % len(d)) # had some trouble with size()
@@ -198,7 +208,8 @@ def strip_leading_slash(filename):
   return filename
 
 def export_commit(ui,repo,revision,old_marks,max,count,authors,
-                  branchesmap,sob,brmap,hgtags,encoding='',fn_encoding='',filter_contents=None):
+                  branchesmap,sob,brmap,hgtags,encoding='',fn_encoding='',filter_contents=None,
+                  plugins={}):
   def get_branchname(name):
     if brmap.has_key(name):
       return brmap[name]
@@ -211,6 +222,16 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
   branch=get_branchname(branch)
 
   parents = [p for p in repo.changelog.parentrevs(revision) if p >= 0]
+  author = get_author(desc,user,authors)
+
+  if plugins and plugins['commit_message_filters']:
+    commit_data = {'branch': branch, 'parents': parents, 'author': author, 'desc': desc}
+    for filter in plugins['commit_message_filters']:
+      filter(commit_data)
+    branch = commit_data['branch']
+    parents = commit_data['parents']
+    author = commit_data['author']
+    desc = commit_data['desc']
 
   if len(parents)==0 and revision != 0:
     wr('reset refs/heads/%s' % branch)
@@ -218,7 +239,7 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
   wr('commit refs/heads/%s' % branch)
   wr('mark :%d' % (revision+1))
   if sob:
-    wr('author %s %d %s' % (get_author(desc,user,authors),time,timezone))
+    wr('author %s %d %s' % (author,time,timezone))
   wr('committer %s %d %s' % (user,time,timezone))
   wr('data %d' % (len(desc)+1)) # wtf?
   wr(desc)
@@ -259,8 +280,8 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
   removed=[strip_leading_slash(x) for x in removed]
 
   map(lambda r: wr('D %s' % r),removed)
-  export_file_contents(ctx,man,added,hgtags,fn_encoding,filter_contents)
-  export_file_contents(ctx,man,changed,hgtags,fn_encoding,filter_contents)
+  export_file_contents(ctx,man,added,hgtags,fn_encoding,filter_contents,plugins)
+  export_file_contents(ctx,man,changed,hgtags,fn_encoding,filter_contents,plugins)
   wr()
 
   return checkpoint(count)
@@ -396,7 +417,8 @@ def verify_heads(ui,repo,cache,force,branchesmap):
 
 def hg2git(repourl,m,marksfile,mappingfile,headsfile,tipfile,
            authors={},branchesmap={},tagsmap={},
-           sob=False,force=False,hgtags=False,notes=False,encoding='',fn_encoding='',filter_contents=None):
+           sob=False,force=False,hgtags=False,notes=False,encoding='',fn_encoding='',filter_contents=None,
+           plugins={}):
   def check_cache(filename, contents):
     if len(contents) == 0:
       sys.stderr.write('Warning: %s does not contain any data, this will probably make an incremental import fail\n' % filename)
@@ -438,7 +460,8 @@ def hg2git(repourl,m,marksfile,mappingfile,headsfile,tipfile,
   brmap={}
   for rev in range(min,max):
     c=export_commit(ui,repo,rev,old_marks,max,c,authors,branchesmap,
-                    sob,brmap,hgtags,encoding,fn_encoding,filter_contents)
+                    sob,brmap,hgtags,encoding,fn_encoding,filter_contents,
+                    plugins)
   if notes:
     for rev in range(min,max):
       c=export_note(ui,repo,rev,c,authors, encoding, rev == min and min != 0)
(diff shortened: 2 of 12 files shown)


frej:
I accept that a 5-10x slowdown is pretty significant, but I'm also very hesitant to add arbitrary conversions, as if I start, I'll have a hard time refusing to add new ones. It is my belief that fast-export helps more users by providing generic mechanism instead of specialized solutions. Why not investigate a Python plugin-based system which could replace the existing filter (which could provide the current functionality using a plugin which runs shell commands)?

I don't get the reason for the second patch. Why would you need it as branches and branch names are preserved during conversion, what is lost without it? Please explain.

jhenkens:
Fair enough - a plugin based system does indeed sound like the proper solution for the dos2unix.

Regarding the branch name in commit - current branches are maintained yes, but mercurial has a concept of "what branch was this commit created on", which git doesn't have a corresponding piece of information. I.e., In git you can commit on an untracked head, which you cannot do in mercurial.

In my case, the branch names of merged-long-ago branches of my repository had much more meaningful information than the commits themselves, and preserving this information was important to me.

frej:
@jhenkens : I see, I didn't know that hg tracked that meta data, interesting.

Why not use git notes for this (we already do that to preserve the hg hash with the --hg-hash flag)? That way it could easily be available to tools, and we don't need to mess with the commit subject (which may be problematic if it overruns the displayed width in various tools). Such a design would require us to decide if we want a new name space (`hg-branch`?) or if we should stay in the `hg`-name space and change the data format to a list of `<key> = <value>` fields (we would still have to provide a way to generate the legacy format).

jhenkens:
I think that both of my improvements would do well as plugins, as said initially for dos2unix, actually. 

Having the branch name as part of the commit message was what we wanted for our use - to have it immediately available in git show and whatnot. 

But I could also see people who would like it in a git note. 

I’ll see if I can figure something out for plugins. Any thoughts?

Sent from my mobile

_________________
Johan Henkens

frej:
A generic plugin mechanism that can be used for both massaging commit messages and files sounds very tempting.

Ideally plugins should should not require external dependencies (check if there is something already in mercurial we could use) and allow us to migrate to Python 3 when/if Mercurial ends support for Python 2.

hg-fast-export.py
```diff
@@ -214,10 +207,10 @@ def get_branchname(name):
...
   parents = [p for p in repo.changelog.parentrevs(revision) if p >= 0]
   author = get_author(desc,user,authors)
-  if branch_name_in_commit == 'start':
-    desc = branch + '\n' + desc
-  if branch_name_in_commit == 'end':
-    desc = desc + '\n' + branch
+
+  for plugin in plugins:
+    for filter in plugin['commit_message_filters']:
+      desc = filter(desc,branch,author)
```
frej: When we in the future discover that we need to send more than these three values to the plugins,we'll have to update the signature for all plugins. Perhaps we should give it a `{desc: <>, branch: <>, author: <> }` dictionary instead?

plugins/shell_filter_file_contents/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+        try:
+            filter_proc = subprocess.Popen(filter_cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+            d, _ = filter_proc.communicate(d)
+        except:
+            sys.stderr.write('Running filter-contents %s:\n' % filter_cmd)
+            raise
+        filter_ret = filter_proc.poll()
+        if filter_ret:
+            raise subprocess.CalledProcessError(filter_ret, filter_cmd)
+        return d
```
frej: no newline

hg-fast-export.py
jhenkens: Ooh. This idea is great. Leave the signature fixed.

jhenkens:
I've updated the MR, but left it as two commits for the time being. Once provisioinally approved I'll squash into one with a cleaner message.

Let me know what you think.

README.md
```diff
@@ -99,27 +99,34 @@ name the -B and -T options allow a mapping file to be specified to
...
-The filter is supplied to the --filter-contents option. hg-fast-export
-runs the filter for each exported file, pipes its content to the filter's
-standard input, and uses the filter's standard output in place
-of the file's original content. The prototypical use of this feature
-is to convert line endings in text files from CRLF to git's preferred LF:
+hg-fast-export supports plugins to manipulate the file data and commit
+metadata. The plugins are enabled with the --plugin option. The value
+of said option is a plugin name (by folder in the plugins directory),
+and optionally, and equals-sign followed by an initialization string. 
+
```
frej: I didn't see that there was a way to pass parameters to the plugin in my first reading and was going to suggest it now, nice work!

README.md
```diff
@@ -99,27 +99,34 @@ name the -B and -T options allow a mapping file to be specified to
...
+There is a readme accompanying each of the bundled plugins, with a 
+description of their usage. To create a new plugin, one must simply
+add a new folder under the `plugins` directory, with the name of their
```
frej: s/their/the/

README.md
```diff
@@ -99,27 +99,34 @@ name the -B and -T options allow a mapping file to be specified to
...
+new plugin. Inside, there must be an `__init__.py` file, which contains
+at a minimum:
```
frej: For distributions which package fast-export (at least Debian, Gentoo and Homebrew does) and install into system directories we should perhaps add a command line flag to add an external directory to the search path. That way you can use the installed fast-export with a custom plugin.

hg-fast-export.py
```diff
@@ -551,23 +561,15 @@ def bail(parser,opt):
       name, opts = split[0], '='.join(split[1:])
       i = [p for p in all_plugins if p["name"] == name]
       if not i:
-        print 'Could not find plugin ' + name
+        sys.stderr.write('Could not find plugin ' + name)
```
frej: Status messages to stderr, I like it.

README.md
```diff
@@ -99,27 +99,34 @@ name the -B and -T options allow a mapping file to be specified to
...
+class Filter:
+    def __init__(self, args):
+        pass
+        #Or don't pass, if you want to do some init code here
 ```
 
+Beyond the boilerplate initialization, you can see the two different 
+defined filter methods in the [dos2unix](./plugins/dos2unix) and 
+[branch_name_in_commit](./plugins/branch_name_in_commit) plugins.
+
```
frej: Please expand on the filter methods, i.e. explain what they are used for, their signature and what they expect in their input dict.

plugins/branch_name_in_commit/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+def build_filter(args):
+    return Filter(args)
+
+class Filter:
+    def __init__(self, args):
+        if not args in ['start','end']:
+            raise Exception('Cannot have branch name anywhere but start and end')
+        self.pos = args
+
+    def commit_message_filter(self,hash):
```
frej: `hash` is the type of the argument, call it `commit_data` or something more descriptive. (this goes for all the filters)

README.md
```diff
@@ -120,6 +120,56 @@ if [ "$3" == "1" ]; then cat; else dos2unix; fi
...
+Beyond the boilerplate initialization, you can see the one of the
+defined filter methods in the [dos2unix](./plugins/dos2unix) plugin.
+
+```
+commit_data = {'branch': branch, 'parents': parents, 'author': author, 'desc': desc}
+
+def commit_message_filter(self,commit_data):
+```
+The `commit_message_filter` method is called for each commit, after parsing 
+from hg, but before outputting to git. The hash `commit_data` contains the
```
frej: _Hash_ is Perl (at least) they are called _dictionaries_ in Python.

pluginloader/__init__.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
```
frej: This is not something we run stand-alone, why have it there at all? If it was for stand-alone use, it should say `python2` to conform to PEP 394.

plugins/dos2unix/__init__.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python2
```
frej: Drop?

plugins/dos2unix/README.md
```diff
@@ -0,0 +1,9 @@
+## Dos2unix filter
```
frej: Sorry I forgot about this part, I should have been clearer in my previous review comment. Could you break out the dos2unix-part to its own commit? That way we get a clean _mechanisms_-commit and then a separate commit for each new feature.

hg-fast-export.py
```diff
@@ -566,19 +554,21 @@ def bail(parser,opt):
...
   if options.filter_contents!=None:
-    import shlex
-    filter_contents=shlex.split(options.filter_contents)
+    plugins+=['shell_filter_file_contents='+options.filter_contents]
 
   plugins={}
   plugins['commit_message_filters']=[]
   plugins['file_data_filters']=[]
 
-  if options.plugins and options.pluginpath:
```
frej: Apart from the filter_contents part, this belongs to the previous _mechanisms_ commit.

hg-fast-export.py
```diff
@@ -566,19 +554,21 @@ def bail(parser,opt):
...
+  if plugins and options.pluginpath:
     sys.stderr.write('Using additional plugin path: ' + options.pluginpath + '\n')
 
-  for plugin in options.plugins:
+  for plugin in plugins:
```
frej: As above

plugins/branch_name_in_commit/__init__.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python2
```
frej: Drop?

jhenkens:
I think I've split it out as requested, and made the necessary changes again. Sorry about the back and forth.

frej:
Don't worry about it, I could have noticed those things earlier. Each time I ask for changes, I worry about you getting fed up and abandoning the patch.

hg-fast-export.sh
```diff
@@ -58,6 +58,8 @@ Options:
 	--mappings-are-raw Assume mappings are raw <key>=<value> lines
 	--filter-contents <cmd>  Pipe contents of each exported file through <cmd>
 	                         with <file-path> <hg-hash> <is-binary> as arguments
+	--plugin <plugin=init>  Add a plugin with the given init string (repeatable)
+    --plugin-path <plugin-path> Add an additional plugin lookup path
```
frej: The indentation is wrong (I remember this from the previous version, but it looks like I did manage to lose the comment)

plugins/shell_filter_file_contents/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    return Filter(args)
+
+class Filter:
+    def __init__(self, args):
+        self.filter_contents = shlex.split(args)
+
+    def file_data_filter(self,file_data):
+        d = file_data['data']
+        file_ctx = file_data['file_ctx']
+        filter_cmd = filter_contents + [filename, node.hex(file_ctx.filenode()), '1' if file_ctx.isbinary() else '0']
```
frej: Have you tried this?

```
  File "/home/frej/src/fast-export/hg-fast-export.py", line 586, in <module>
    plugins=plugins_dict))
  File "/home/frej/src/fast-export/hg-fast-export.py", line 452, in hg2git
    plugins)
  File "/home/frej/src/fast-export/hg-fast-export.py", line 271, in export_commit
    export_file_contents(ctx,man,added,hgtags,fn_encoding,plugins)
  File "/home/frej/src/fast-export/hg-fast-export.py", line 145, in export_file_contents
    filter(file_data)
  File "/home/frej/src/fast-export/pluginloader/../plugins/shell_filter_file_contents/__init__.py", line 15, in file_data_filter
    filter_cmd = filter_contents + [filename, node.hex(file_ctx.filenode()), '1' if file_ctx.isbinary() else '0']
NameError: global name 'filter_contents' is not defined
```

`filename` and `node` also look suspicious.

jhenkens: My mistake  - I think I may have had it working in the past, and lost it in the various reworks, but I can't be certain. I have fixed this (and you were right, filename and node are also suspicious, as was sys) and verified that all three plugins are currently working as intended.

jhenkens:
Teaches me for even attempting it on WSL. 😉 
I've gone ahead and removed the trailing whitespace and done yet another pass of `dos2unix`, from a proper unix box for sanity.

hg-fast-export.py
```diff
@@ -211,14 +210,24 @@ def get_branchname(name):
...
+
+  if plugins and plugins['commit_message_filters']:
+    commit_data = {'branch': branch, 'parents': parents, 'author': author, 'desc': desc}
+    for filter in plugins['commit_message_filters']:
+      filter(commit_data)
+    branch = commit_data['branch']
+    parents = commit_data['parents']
+    author = commit_data['author']
+    desc = commit_data['desc']
+
```
frej: trailing whitespace

plugins/branch_name_in_commit/README.md
```diff
@@ -0,0 +1,10 @@
+## Branch Name in Commit Message
+
+Mercurial has a much stronger notion of branches than Git,
```
frej: trailing whitespace

plugins/branch_name_in_commit/README.md
```diff
@@ -0,0 +1,10 @@
+## Branch Name in Commit Message
+
+Mercurial has a much stronger notion of branches than Git, 
+and some parties may not wish to lose the branch information
```
frej: here too

plugins/branch_name_in_commit/README.md
```diff
@@ -0,0 +1,10 @@
+## Branch Name in Commit Message
+
+Mercurial has a much stronger notion of branches than Git, 
+and some parties may not wish to lose the branch information 
+during the migration to Git. You can use this plugin to either
+prepend or append the branch name from the mercurial
+commit into the commit message in Git.
+
+To use the plugin, add
```
frej: this line too

plugins/branch_name_in_commit/README.md
```diff
@@ -0,0 +1,10 @@
+## Branch Name in Commit Message
+
+Mercurial has a much stronger notion of branches than Git, 
+and some parties may not wish to lose the branch information 
+during the migration to Git. You can use this plugin to either
+prepend or append the branch name from the mercurial
+commit into the commit message in Git.
+
+To use the plugin, add 
+`--plugin branch_name_in_commit=(start|end)`.
```
frej: this one too

plugins/dos2unix/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+This plugin converts CRLF line ending to LF in text files in the repo.
+It is recommended that you have a .gitattributes file that maintains
```
frej: here too

plugins/dos2unix/README.md
```diff
@@ -0,0 +1,9 @@
+## Dos2unix filter
+
+This plugin converts CRLF line ending to LF in text files in the repo.
+It is recommended that you have a .gitattributes file that maintains 
+the usage of LF endings going forward, for after you have converted your
```
frej: and here

plugins/dos2unix/README.md
```diff
@@ -0,0 +1,9 @@
+## Dos2unix filter
+
+This plugin converts CRLF line ending to LF in text files in the repo.
+It is recommended that you have a .gitattributes file that maintains 
+the usage of LF endings going forward, for after you have converted your 
+repository.
+
+To use the plugin, add
```
frej: here

plugins/dos2unix/README.md
```diff
@@ -0,0 +1,9 @@
+## Dos2unix filter
+
+This plugin converts CRLF line ending to LF in text files in the repo.
+It is recommended that you have a .gitattributes file that maintains 
+the usage of LF endings going forward, for after you have converted your 
+repository.
+
+To use the plugin, add 
+`--plugin dos2unix`.
```
frej: and here

plugins/shell_filter_file_contents/README.md
```diff
@@ -0,0 +1,30 @@
+## Shell Script File Filter
+
+This plugin uses shell scripts in order to perform filtering of files.
+If your preferred scripting is done via shell, this tool is for you.
+Be noted, though, that this method can cause an order of magnitude slow
```
frej: here

plugins/shell_filter_file_contents/README.md
```diff
@@ -0,0 +1,30 @@
+## Shell Script File Filter
+
+This plugin uses shell scripts in order to perform filtering of files.
+If your preferred scripting is done via shell, this tool is for you.
+Be noted, though, that this method can cause an order of magnitude slow 
+down. For small repositories, this wont be an issue.
+
+To use the plugin, add
```
frej: here

plugins/shell_filter_file_contents/README.md
```diff
@@ -0,0 +1,30 @@
+## Shell Script File Filter
+
+This plugin uses shell scripts in order to perform filtering of files.
+If your preferred scripting is done via shell, this tool is for you.
+Be noted, though, that this method can cause an order of magnitude slow 
+down. For small repositories, this wont be an issue.
+
+To use the plugin, add 
+`--plugin shell_filter_file_contents=path/to/shell/script.sh`.
```
frej: here

plugins/shell_filter_file_contents/README.md
```diff
@@ -0,0 +1,30 @@
+## Shell Script File Filter
+
+This plugin uses shell scripts in order to perform filtering of files.
+If your preferred scripting is done via shell, this tool is for you.
+Be noted, though, that this method can cause an order of magnitude slow 
+down. For small repositories, this wont be an issue.
+
+To use the plugin, add 
+`--plugin shell_filter_file_contents=path/to/shell/script.sh`. 
+The filter script is supplied to the plugin option after the plugin name,
```
frej: here

frej:
Now it looks good. Thank you for your contribution and patience.

PR #157, opened 2019-07-25
Remove git submodules if .hgsubstate file was removed or emptied

### LINKED ISSUE #156: Git submodule still appears after hg subrepo have been removed
**Steps to reproduce:**
- Create hg subrepo
- Remove hg subrepo
- Convert repository with `fast-export`

**Expected behavior:**
No git submodule in head commit

**Actual behavior:**
Subrepository still there

MokhamedDakhraui:
Fix removal of git subrepo if it's been deleted as hg subrepo

Fixes #156

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -313,12 +313,12 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
   sys.stderr.write('%s: Exporting %s revision %d/%d with %d/%d/%d added/changed/removed files\n' %
       (branch,type,revision+1,max,len(added),len(changed),len(removed)))
 
-  if fn_encoding:
-    removed=[r.decode(fn_encoding).encode('utf8') for r in removed]
+  for filename in removed:
+    if fn_encoding:
+      filename=filename.decode(fn_encoding).encode('utf8')
+    filename=strip_leading_slash(filename)
+    wr('D %s' % filename)
 
-  removed=[strip_leading_slash(x) for x in removed]
-
-  map(lambda r: wr('D %s' % r),removed)
   export_file_contents(ctx,man,added,hgtags,fn_encoding,plugins)
   export_file_contents(ctx,man,changed,hgtags,fn_encoding,plugins)
   wr()
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -128,11 +128,14 @@ def get_author(logmessage,committer,authors):
   return committer
 
 def remove_gitmodules(ctx):
-  """Removes all submodules"""
-  # Remove all submodules as we don't detect deleted submodules properly
-  # in any other way. We will add the ones not deleted back again below.
-  for module in submodule_mappings.keys():
-    wr('D %s' % module)
+  """Removes all submodules of ctx parents"""
+  # Removing all submoduies coming from all parents is safe, as the submodules
+  # of the current commit will be re-added below. A possible optimization would
+  # be to only remove the submodules of the first parent.
+  for parent_ctx in ctx.parents():
+    for submodule in parent_ctx.substate.keys():
+      wr('D %s' % submodule)
+  wr('D .gitmodules')
 
 def refresh_gitmodules(ctx):
   """Updates list of ctx submodules according to .hgsubstate file"""
@@ -171,7 +174,7 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding='',plugins={}):
   count=0
   max=len(files)
   for file in files:
-    if submodule_mappings and ctx.substate and file==".hgsubstate":
+    if submodule_mappings and file==".hgsubstate":
       refresh_gitmodules(ctx)
     # Skip .hgtags files. They only get us in trouble.
     if not hgtags and file == ".hgtags":
@@ -311,6 +314,8 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
     if fn_encoding:
       filename=filename.decode(fn_encoding).encode('utf8')
     filename=strip_leading_slash(filename)
+    if filename=='.hgsubstate':
+      remove_gitmodules(ctx)
     wr('D %s' % filename)
 
   export_file_contents(ctx,man,added,hgtags,fn_encoding,plugins)
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -127,52 +127,58 @@ def get_author(logmessage,committer,authors):
       return r
   return committer
 
+def remove_gitmodules(ctx):
+  """Removes all submodules"""
+  # Remove all submodules as we don't detect deleted submodules properly
+  # in any other way. We will add the ones not deleted back again below.
+  for module in submodule_mappings.keys():
+    wr('D %s' % module)
+
+def refresh_gitmodules(ctx):
+  """Updates list of ctx submodules according to .hgsubstate file"""
+  remove_gitmodules(ctx)
+  # Read .hgsubstate file in order to find the revision of each subrepo
+  data=ctx.filectx(".hgsubstate").data()
+  subHashes={}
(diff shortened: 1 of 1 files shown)


hg-fast-export.py
```diff
@@ -164,6 +168,7 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding='',plugins={}):
           wr('M 160000 %s %s' % (gitSha, name))
           sys.stderr.write("Adding submodule %s, revision %s->%s\n"
                            % (name,subHashes[name],gitSha))
+
```
frej: Spurious white-space change, please remove.

hg-fast-export.py
```diff
@@ -127,16 +127,20 @@ def get_author(logmessage,committer,authors):
...
   max=len(files)
-  for file in files:
-    if submodule_mappings and ctx.substate and file==".hgsubstate":
-      # Remove all submodules as we don't detect deleted submodules properly
-      # in any other way. We will add the ones not deleted back again below.
-      for module in submodule_mappings.keys():
-        wr('D %s' % module)
 
+  for file in files:
+    if submodule_mappings and file==".hgsubstate":
```
frej: Can you explain why you drop `ctx.substate`?

hg-fast-export.py
```diff
@@ -311,6 +316,9 @@ def get_branchname(name):
     removed=[r.decode(fn_encoding).encode('utf8') for r in removed]
 
   removed=[strip_leading_slash(x) for x in removed]
+  
+  if '.hgsubstate' in added+removed+changed:
```
frej: This check seems like it could lead to a significant slowdown. You are scanning through all changed files for each commit. As you know what the file name is, you should be able to look it up in the manifest (which uses dictionaries) and determine if it has changed in O(1) instead of O(N).

hg-fast-export.py
MokhamedDakhraui: Ah, it's residual change from my first attempt to locate and fix this bug. Previously after removal of only subrepo it prevented update of `.gitmodules` file.

Now that isn't have a meaning, so I'll revert it back.

hg-fast-export.py
MokhamedDakhraui: Yeah, that was quite crude fix. I tried to make as little changes in code as possible due to my lack of knowledge in python and, I suppose, internals of mercurial.

I'll try to use manifest, but it'll require more thorough review :)

frej:
@MokhamedDakhraui: What's the status of this work? If you're done I'd prefer if you would squash this patch series to a single commit.

MokhamedDakhraui:
> @MokhamedDakhraui: What's the status of this work? If you're done I'd prefer if you squash this patch series to a single commit.

It's done

MokhamedDakhraui:
I made error in patch, will fix it soon

hg-fast-export.py
```diff
@@ -127,52 +127,56 @@ def get_author(logmessage,committer,authors):
...
-        wr('D %s' % module)
-
-      # Read .hgsubstate file in order to find the revision of each subrepo
-      data=ctx.filectx(file).data()
-      subHashes={}
-      for line in data.split('\n'):
-        if line.strip()=="":
-          continue
-        cols=line.split(' ')
-        subHashes[cols[1]]=cols[0]
```
MokhamedDakhraui: Removed parsing of `.hgsubstate` since `ctx.substate` already has all required info

MokhamedDakhraui:
Completed fix and ready to review.
Bug had been appeared in next case:
- one of subrepositories, say `FishRepo`, had been removed;
- farther in history created `FishRepo` folder
- changed substate of another subrepository

After subrepository state had been changed all folders with names listed in file passed with `--subrepos` parameter will be removed (in resulted git repository).

https://github.com/frej/fast-export/blob/47d330de83055efc5d4eeec6d0593d7aa3640555/hg-fast-export.py#L137-L138

frej:
This looks reasonable, but please don't mix in the submodule stuff in the refactoring. Create two commits: The first one refactors out the submodule stuff from `export_file_contents` and `export_commit`, the second updates the newly created `remove_gitmodules` and `export_file_contents`. That way the first commit will be obviously correct (tm), and lazy me will only have to think about the second part :)

MokhamedDakhraui:
Reasonable enough. Will do that.

Previous message about squash and structure of other PRs from this repo formed wrong impression that requirement is "one PR - one commit". Glad that is not the case)

frej:
One commit per logical change is the main rule. When patching a commit due to review comments I prefer an amended commit as I won't have to apply the original patch, apply the update and then diff against master in my head when reviewing the change.

MokhamedDakhraui:
Ok, finished.
Divided to 3 parts:
- simple move code
- refactor
- actual fix
Hopefully, you enjoy the review :)

hg-fast-export.py
```diff
@@ -128,11 +128,13 @@ def get_author(logmessage,committer,authors):
...
 
 def remove_gitmodules(ctx):
-  """Removes all submodules"""
-  # Remove all submodules as we don't detect deleted submodules properly
-  # in any other way. We will add the ones not deleted back again below.
-  for module in submodule_mappings.keys():
-    wr('D %s' % module)
+  """Removes all submodules of ctx parents"""
+  # Not quite sure whether submodules from all parents should be removed.
+  # Subrepos from first parent may be sufficient.
```
frej: How does the earlier version's parsed `.hgsubstate` file differ from using all the parents? Is it equivalent? Have you analyzed this?

hg-fast-export.py
frej: I don't get it, why would a check of `ctx.substate` trigger parsing of the `.hgsubstate` file, if `submodule_mappings` is true, we will already have accessed it early in hg2git()? Please explain.

MokhamedDakhraui: For clarification:
Algorithm of updating list of submodules (when `.hgsubstate` file'd been added/modified) implemented as follows:
- removed all parents' subrepos
- added subrepos from current revision
---
If `.hgsubstate` file was emptied it can be recognized as modified (theoretically), but `ctx.substate` will be empty, and list of submodules won't be updated.
In fact I tested that case, and, wierdly, empty `.hgsubstate` had been considered as removed, but not modified. It might be related to "apology" comment about shortcut in `status` function, but for now I did't investigate it further. My point is: empty and changed substate doesn't mean that submodule list shouldn't be updated.

About overhead of parsing `.hgsubstate` file on `ctx.substate` call, this is a bit of ignorance from my side. I looked again in function `subrepoutil.state` and found out that parsed state is cached.

MokhamedDakhraui: If you think the argument is not sufficient, I will add substate check back (it shouldn't break anything anyway, I suppose)

MokhamedDakhraui:
> 
> 
> That the second line of the commit message starts with "Also" is a red flag that this commit should be broken into two. Otherwise, LGTM.

Hm, good idea. Will fix

hg-fast-export.py
frej: I buy your argument, good!

hg-fast-export.py
frej: From your reasoning it is clear that removing the submodules from all parents should be safe, make the comment more assertive so the next contributor won't have analyze the code again. I'm thinking: "Removing all submoduies coming from all parents is safe, as the submodules of the current commit will be re-added below. A possible optimization would be to only remove the submodules of the first parent"

frej: Please line break the comment at 80 columns. The comment, as written in c1089b015442ae2ae3834bcc3a2147036a7b93ff, is not grammatically correct, why not simply use what I suggested above?

frej:
And we're done. Thank you for your contribution!

Issue #143, opened 2018-08-31
./hg-fast-export.sh: line 177: python2: command not found


GregWoods:
Although it's likely this is a configuration issue, I am stuck

- Windows 10 64 bit
- Python 2.7.15 installed and in the path
- Mercurial 4.7 install, and in the path
- Git 2.18.0.windows.1 installed, and in the path

Running the following command in git-bash:

`./hg-fast-export.sh -r ../pathToHgRepo`

Looking at hg-fast-export.sh, line 177 is a comment, so not sure where to go from here.

Any help appreciated.

Troublem4ker:
I had the same issue. For me a downgrade to tag/v180317 fixed this problem.

klm1:
> I had the same issue. For me a downgrade to tag/v180317 fixed this problem.

Same here -- downgraded version works.

AlexP11223:
Yeah, looks like there is no python2 in Windows Python distributions.

klm1:
Same issue on macOS 10.13.

andres-carrilloAdapt2:
How did you Downgrade?

klm1:
I used:
git checkout tags/v180317

AlexP11223:
You can download it here. https://github.com/frej/fast-export/tree/eca99b61ebb1fa5a2eb2e33a80d05212f5af7539

Or clone and checkout using git.

sikmir:
https://github.com/frej/fast-export/blob/47d330de83055efc5d4eeec6d0593d7aa3640555/hg-fast-export.sh#L29
Just run as:
```bash
$ PYTHON=python ./hg-fast-export.sh -r ../pathToHgRepo
```

pdatta1988:
i could not resolve the issue of python2. If i give PYTHON=python i am getting below error 👎
fast-export/hg-fast-export.py", line 6, in <module>
    from mercurial import node
ImportError: No module named mercurial


Please suggest a solution.

AlexP11223:
> No module named mercurial

it's a different error. You probably don't have Mercurial installed (or it's not in your PATH for some reason).

jkfindeisen:
> Please suggest a solution.

It means you got past that error here and arrived at the next one on Windows. Please make sure you have a Python 2.X with the mercurial module installed running.

njackowski:
Not sure how helpful this is, but I just ran into the same thing.  
Windows 10 x64
Python 2.7.16

[...]

Issue #160, opened 2019-08-27
fast-export/hg-fast-export.sh: line 179: python2: command not found 


VitaliyAT:
I can't convert any hg to git and got line as in title, 179 line it's a comment "# move recent marks cache out of the way..."
I have both pythons installed: 2.7.16 and 3.7.3 Also Win 10

frej:
I can guess that the line number is your shell being confused about the redirect/exec magic ending at line 179. Is `python2` in you path? #143 looks suspiciously like this issue.

clach04:
The workaround of setting PYTHON to python worked for me on my Microsoft Windows machine but applying the code fix is easier.

I.e.

    env PYTHON=python hg-fast-export.sh  ....

frej:
Let me known if the patch in f57fba000b9f15ebcb68e898db23ef02a23eaefe doesn't fix your problem @VitaliyAT.

@clach04: Regarding your patch in 162ba4a3f2cb3c977917e3e270720fe9ce10ca7e: Checking for a particular OS and then assuming that a particular feature behaves in a certain way is known to not be robust. It's much better to check the feature instead, as that will do the right thing on any OS or platform. In this case the patch in f57fba000b9f15ebcb68e898db23ef02a23eaefe will (hopefully) work both on windows and Arch-linux (the reason for [PEP 394](https://www.python.org/dev/peps/pe)).

PR #162, opened 2019-08-31
Resolve issues #143 and #160

clach04:
Microsoft Windows distributions of Python do not ship a python2.exe.
Added a simply check for msys. Considered `uname -s` but this includes
both platform bit-ness and Windows OS version so is a moving target.

### DIFF
--- a/hg-fast-export.sh
+++ b/hg-fast-export.sh
@@ -26,7 +26,12 @@ SFX_MARKS="marks"
 SFX_HEADS="heads"
 SFX_STATE="state"
 GFI_OPTS=""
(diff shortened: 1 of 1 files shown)


frej:
Checking for a particular OS and then assuming that a particular feature behaves in a certain way is known to not be robust. It's much better to check the feature instead, as that will do the right thing on any OS or platform. In this case the patch in f57fba000b9f15ebcb68e898db23ef02a23eaefe will (hopefully) work both on windows and Arch-linux (the reason for [PEP 394](https://www.python.org/dev/peps/pe)).

PR #167, opened 2019-09-13
Run 2to3 on project files

dblanchette:
This does not add Python 3 compatibility because mercurial still does not work with Python 3 (at least according to my tests with mercurial 5.1.1).

However, the changes in this PR do not break Python 2 compatibility and makes the code of fast-export itself compatible with Python 3.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -193,8 +193,8 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding='',plugins={}):
 
     if plugins and plugins['file_data_filters']:
       file_data = {'filename':filename,'file_ctx':file_ctx,'data':d}
-      for filter in plugins['file_data_filters']:
-        filter(file_data)
+      for filter_fn in plugins['file_data_filters']:
+        filter_fn(file_data)
       d=file_data['data']
       filename=file_data['filename']
       file_ctx=file_data['file_ctx']
@@ -253,7 +253,7 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
                   branchesmap,sob,brmap,hgtags,encoding='',fn_encoding='',
                   plugins={}):
   def get_branchname(name):
-    if brmap.has_key(name):
+    if name in brmap:
       return brmap[name]
     n=sanitize_name(name, "branch", branchesmap)
     brmap[name]=n
@@ -268,8 +268,8 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
 
   if plugins and plugins['commit_message_filters']:
     commit_data = {'branch': branch, 'parents': parents, 'author': author, 'desc': desc}
-    for filter in plugins['commit_message_filters']:
-      filter(commit_data)
+    for filter_fn in plugins['commit_message_filters']:
+      filter_fn(commit_data)
     branch = commit_data['branch']
     parents = commit_data['parents']
     author = commit_data['author']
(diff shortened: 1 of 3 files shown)


frej:
I prefer to defer the 2 to 3 conversion until  Mercurial has released a supported version with Python 3 support, so I won't merge this now. 

When you say that it doesn't work with Python 3, is this out of the box or are you using the work-arounds described on [Mercurial's Python3 page]( https://www.mercurial-scm.org/wiki/Python3)? Also I would prefer if you tried to coordinate your work with the contributors on #90 as I think @ralic i a Mercurial developer.

dblanchette:
@frej I made this mostly because there was no activity on #90 and I could not go there and fix the issues. You may ignore me and work on the other one, I won't mind. However, the changes in this PR are compatible with Python 2 and 3 so it does not break anything, just makes it easier to convert when need be.

To test with Python 3, I did the following:

- Create a virtual environment using Python 3.7.4
- Installed Mercurial with `HGPYTHON3=1 HGALLOWPYTHON3=1 pip install mercurial`
- In a git bash window (I'm on Windows), ran `PYTHONLEGACYWINDOWSFSENCODING=1 hg-fast-export.sh -r /path/to/a/mercurial/repo`

I got a stack trace:
```Traceback (most recent call last):
  File "D:\Repos\hg-fast-export\hg2git.py", line 34, in setup_repo
    myui=ui.ui(interactive=False)
TypeError: __init__() got an unexpected keyword argument 'interactive'

During handling of the above exception, another exception occurred:

  File "D:/Repos/hg-fast-export/hg-fast-export.py", line 665, in <module>
    plugins=plugins_dict))
  File "D:/Repos/hg-fast-export/hg-fast-export.py", line 483, in hg2git
    ui,repo=setup_repo(repourl)
  File "D:\Repos\hg-fast-export\hg2git.py", line 37, in setup_repo
    myui.setconfig('ui', 'interactive', 'off')
  File "D:\Repos\hg-fast-export\py37_venv\lib\site-packages\mercurial\ui.py", line 517, in setconfig
    cfg.set(section, name, value, source)
  File "D:\Repos\hg-fast-export\py37_venv\lib\site-packages\mercurial\config.py", line 82, in set
    'config section may not be unicode strings on Python 3')
AssertionError: b'config section may not be unicode strings on Python 3'```

junghans:
As Fedora is removing python2 in the next release - any news on this?

ralic:
This version of mercury shall be working in python3.
https://github.com/ralic/mercurial_py3

junghans:
> This version of mercury shall be working in python3.
> https://github.com/ralic/mercurial_py3

Mercurial 5.2 support python3 as well: https://www.mercurial-scm.org/wiki/Python3

chrisjbillington:
Porting hg-fast-export to python 3 is going to be considerably more involved than 2to3, because mercurial has made the IMHO extremely misguided decision to keep everything in the mercurial API bytes-only.

Porting will have to basically convert things to bytes for interacting with the mercurial API, keeping them otherwise stringy within hg-fast-export.py so that they work with Python string methods and stdlib functions, and so that they print correctly to stdout. Any code that touches the mercurial API from Python is going to be a royal pain on Python 3.

It's enough to want to wait and convince mercurial to start accepting strings in their API, or to try and use their magic source transformer on hg-fast-export to get it working on Python 3 (though that means maintaining hg-fast-export as Py2 syntax forever, not very appealing!)

frej:
#201 is the way forward, it handles the Mercurial API changes.

Issue #61, opened 2016-02-08
Import HG bookmarks to Git branches


tobia:
HG bookmarks are effectively the same thing as Git feature branches, while HG branches are a very different beast, much less useful IMHO.

Therefore it would be useful if fast-export converted HG bookmarks → Git branches, instead of failing with an error:

    Error: repository has at least one unnamed head: hg r779

frej:
I would not oppose such a feature, feel free to submit a patch.

atifaziz:
:+1: Hg bookmarks is the right call here

HangingClowns:
not using HG to begin with is the real answer :D damn unnamed heads!

tobia:
Here's a workaround to import such a repository.

For each unnamed head, or maybe for each bookmark (`hg bookmarks`) create a HG branch with the same name:

```
hg checkout SomeBookmark
hg branch SomeBookmark
touch .delete_me
hg commit -m SomeBookmark
```

This should be done on the HG repository before trying to import it into Git. You need to change something and make a commit for the new branch to be visible.

chrisjbillington:
You don't need to add a new commit or a new file. You can amend the commit to change the branch it's on:

```bash
hg update SomeBookmark
hg branch SomeBookmark
hg commit --amend -m "$(hg log -r SomeBookmark --template '{desc}')"
```
The `-m` argument is just to copy the commit message from the existing one - if you're interactive you can omit it and just dismiss the editor when it pops up to keep the commit message.

This is a bit cleaner! On the git end it will just look like the branch was always a named branch.

PR #170, opened 2019-09-20
Patch for #61

krivard:
Hello, I ran into the bookmarks / extra heads issue while converting a couple of repos this week, and wrote a small patch to address it -- when building the changeset tuple it checks for bookmarks and, if there are any, appends the first one found to the reported branch name. It does require some cleanup in git before you publish the converted repository, but it worked for us.

### DIFF
--- a/hg2git.py
+++ b/hg2git.py
@@ -5,7 +5,7 @@
 
 from mercurial import hg,util,ui,templatefilters
 from mercurial import error as hgerror
-from mercurial.scmutil import revsymbol,binnode
+from mercurial.scmutil import revsymbol,binnode,revsingle
 
 import re
 import os
@@ -87,6 +87,9 @@ def get_changeset(ui,repo,revision,authors={},encoding=''):
     desc=desc.decode(encoding).encode('utf8')
   tz="%+03d%02d" % (-timezone / 3600, ((-timezone % 3600) / 60))
   branch=get_branch(extra.get('branch','master'))
+  bookmarks = repo.nodebookmarks(node)
+  if bookmarks: branch="%s_%s"%(branch,bookmarks[0])
+  if 'bookmarks' not in extra: extra['bookmakrs'] = bookmarks
   return (node,manifest,fixup_user(user,authors),(time,tz),files,desc,branch,extra)
 
 def mangle_key(key):


hg2git.py
```diff
@@ -5,7 +5,7 @@
 
 from mercurial import hg,util,ui,templatefilters
 from mercurial import error as hgerror
-from mercurial.scmutil import revsymbol,binnode
+from mercurial.scmutil import revsymbol,binnode,revsingle
```
frej: I can't see that this import is used

hg2git.py
```diff
@@ -87,6 +87,9 @@ def get_changeset(ui,repo,revision,authors={},encoding=''):
     desc=desc.decode(encoding).encode('utf8')
   tz="%+03d%02d" % (-timezone / 3600, ((-timezone % 3600) / 60))
   branch=get_branch(extra.get('branch','master'))
+  bookmarks = repo.nodebookmarks(node)
+  if bookmarks: branch="%s_%s"%(branch,bookmarks[0])
+  if 'bookmarks' not in extra: extra['bookmakrs'] = bookmarks
```
frej: `bookmakrs`?

frej:
I pretty much have no clue about the representation of bookmarks in the hg repo. But doesn't this patch silently ignore all but the first bookmark for a change set? Also, as far as I understand, shouldn't we treat bookmarks as branches, with the bookmark name as the branch name? Not a concatenation of the branch and the bookmark? Please explain why your patch is doing "the right thing" as right now I'm not convinced.

krivard:
In hg, branches are permanent metadata on each commit. [Bookmarks](https://www.mercurial-scm.org/wiki/Bookmarks) are lightweight labels for a single commit only, and "follow" the tip of a git-style feature branch. As far as I can tell, there's no way to know where the "start" of a bookmarked git-style branch is in mercurial. Bookmarked commits are also members of some particular hg-style branch, and as soon as a descendant commit is added to the repository, the bookmark is stripped and the permanent branch is the only thing left on the commit record.

Even though the (literally) nominal name for the converted branch is the bookmark, I wasn't comfortable stripping out the permanent branch data completely, since this is sometimes necessary for auditing down the line -- hence the combined name. That's what I mean by "some cleanup in git is required:" if you want to revert back to the bookmark-only branch name for the converted repository, that's easy enough to do. The commit history is maintained, with the tip of the permanent branch remaining where hg says it is, not at commit n-1 of the bookmark branch.

This patch is really just focused on getting fast-export to run, **at all**, on hg repositories that use bookmarks to label multiple heads. If this meets your standard, I can rebase to fix the typo, commit message, and discarded import -- let me know.

frej:
Thanks for the explanation.

> The commit history is maintained, with the tip of the permanent branch remaining where hg says it is, not at commit n-1 of the bookmark branch.

Do you mean that this is what will happen if the branch name is set to the bookmark name, or is it the reason for concatenating the branch and bookmark names? We must maintain the permanent branch so it matches up with what's on the Mercurial side.

What's the plan for multiple bookmarks for the same commit?

Don't put any time into updating the patch until we are clear on what the "right thing" is.

frej:
Thinking further about this, what worries me is what happens if we on the Mercurial side have a repo looking like this (C[0-9]+ are commits):
```
branch master: C0->C1[bookmarked as "foo" which is active ]
```

Then we add a commit so we get:
```
branch master: C0->C1->C2[bookmarked as "foo" which is active ]
```
How will (with the proposed patch) the converted git repo look? Will it turn out as (in beautiful gitk lookalike ascii-art):
```
+ C2 master, master_foo
+ C1
+ C0
```
or
```
+ C2 master_foo
+ C1 master
+ C0
```

krivard:
In that example, there is only one head for the branch, and that head is
bookmarked. Mercurial doesn't store bookmark history, so there is no way to
tell whether 'foo' was carried forward from C1, or was applied for the
first time at C2 for a new feature.

This patch would yield your second converted git repo. Then, depending on
what was actually meant, you'd use git to either drop the master branch and
rename master_foo to master, or, for the case where you started a new
feature foo at C2, you'd leave master where it is and rename master_foo to
foo.

On multiple bookmarks: this patch just takes the first-listed bookmark, so
any remaining bookmarks would be ignored, which I agree isn't great. Does
git permit the same changeset to be present in multiple branches? What part
of the fast-export code would handle that?

On maintenance: "We must maintain the permanent branch so it matches up
with what's on the Mercurial side."  Why is that important, and what
definition of matching is required?

frej:
> This patch would yield your second converted git repo. Then, depending on
what was actually meant, you'd use git to either drop the master branch and
rename master_foo to master, or, for the case where you started a new
feature foo at C2, you'd leave master where it is and rename master_foo to
foo.

This is bad, the manual fixup precludes automated incremental conversions. From a maintainer point of view this behaviour is hopeless to document, defend and explain.

> On multiple bookmarks: this patch just takes the first-listed bookmark, so
any remaining bookmarks would be ignored, which I agree isn't great. Does
git permit the same changeset to be present in multiple branches?

Git does not really work on changesets, a commit is just some metadata plus a snapshot of the file tree. A commit can be a member of multiple branches, that's what happens when you branch. The commits up to the branch point will be a member of both branches.

> What part of the fast-export code would handle that?

You would have to make the loop at line 514 produce a commit for the base branch and also for each head or perhaps this could be implemented more efficiently inside `export_commit()`.

> On maintenance: "We must maintain the permanent branch so it matches up
with what's on the Mercurial side."  Why is that important, and what
definition of matching is required?

If we are running incremental imports things will be very confused if our named branch suddenly is lagging one commit behind (and a nightmare to explain and defend).

krivard:
As the maintainer, you're in the best position to know  the importance of
automated incremental conversions in your current user base. However, you
have zero users who use automated incremental conversions with bookmarks
defining multiple heads, because bookmarks defining multiple heads crashes
fast-export.

The record we get from hg is ambiguous, so it sounds like there are a few
choices here, none of them ideal:

 - Use the current fast-export. Bookmarks are handled incorrectly 100% of
the time.
 - Use the manual fixup. Bookmarks will always work for a single import,
but automated incremental conversions will be broken (for bookmarks users,
and bookmarks users only)
 - Use a heuristic to resolve the ambiguity. In the cases where it guesses
right, bookmarks with automated incremental conversions will work. If it
guesses wrong, it might be tricky for the user to notice. I'm not sure how
bad that would be.

We could go with #2 but make bookmark handling disabled by default and
explicitly incompatible with incremental conversions. We could hedge on #3
by devising some kind of mapping file to store the bookmark branch points
that hg neglects to remember. We could also just say bookmarks are outside
of scope for fast-export, I'll rename my fork, and you can link there from
the bookmarks issue as a possible option for people just doing a single
conversion. Other options I missed?

frej:
> As the maintainer, you're in the best position to know  the importance of
automated incremental conversions in your current user base.

The nasty thing is that I won't know until I break it for them :)

> Other options I missed?

Make bookmarks behave as named branches and make them show up as branches in the `refs/heads/bookmarks/` namespace. As bookmarks behave pretty much as git branches does, we have a natural and understandable mapping. By using a separate namespace we won't get clashes with branch names.

To implement this I think we would get away with:

- Change the detector for unnamed heads to understand bookmarks and ignore them.

- Make `export_commit()` finish by outputting `reset refs/heads/bookmarks/<bookmark>\n from :<mark> ` where `<mark>` is the mark we made on line 282 for each bookmark at this revision.

frej:
A second option would be to just tell the user to convert bookmarks to named branches on the Mercurial side.

frej:
No response in close to a year, closing.

PR #175, opened 2019-10-19
Ignore obsolete changesets in the source repository. Fixes #173

### LINKED ISSUE #173: Unable to convert repository that has been used with the evolve extension
Attempting to convert a changeset that has some obsolete changesets in it just fails with:

```
obsolete feature not enabled but 9 markers found!
  File "/Users/dave/workspace/tmp/fast-export/hg-fast-export.py", line 665, in <module>
    plugins=plugins_dict))
  File "/Users/dave/workspace/tmp/fast-export/hg-fast-export.py", line 499, in hg2git
    (revnode,_,_,_,_,_,_,_)=get_changeset(ui,repo,rev,authors)
  File "/Users/dave/workspace/tmp/fast-export/hg2git.py", line 81, in get_changeset
    node=binnode(revsymbol(repo,str(revision))) # We were given a numeric rev
  File "/usr/local/lib/python2.7/site-packages/mercurial/scmutil.py", line 624, in revsymbol
    raise _filterederror(repo, symbol)
mercurial.error.FilteredRepoLookupError: hidden revision '29' was rewritten as: adf5ac0c33aa
```

Mossop:
This appears to fix the issue at least for my local repository. It switches to an unfiltered repository (the filtered one throws on an attempt to access obsolete revisions) and then filters out the obsolete revisions when it comes across them.

Also sets a simple configuration to avoid the warning about obsolete revisions being present.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -260,6 +260,8 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
     return n
 
   (revnode,_,user,(time,timezone),files,desc,branch,_)=get_changeset(ui,repo,revision,authors,encoding)
+  if repo[revnode].hidden():
+    return count
 
   branch=get_branchname(branch)
 
@@ -332,6 +334,8 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
 
 def export_note(ui,repo,revision,count,authors,encoding,is_first):
   (revnode,_,user,(time,timezone),_,_,_,_)=get_changeset(ui,repo,revision,authors,encoding)
+  if repo[revnode].hidden():
+    return count
 
   parents = [p for p in repo.changelog.parentrevs(revision) if p >= 0]
 
@@ -449,7 +453,7 @@ def verify_heads(ui,repo,cache,force,branchesmap):
 
   # verify that branch has exactly one head
   t={}
-  for h in repo.heads():
+  for h in repo.filtered('visible').heads():
     (_,_,_,_,_,_,branch,_)=get_changeset(ui,repo,h)
     if t.get(branch,False):
       sys.stderr.write('Error: repository has at least one unnamed head: hg r%s\n' %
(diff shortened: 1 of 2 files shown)


Mossop:
It looks like an alternative here would be to avoid filtering over the numeric changeset ids and instead just filter over the filtered repository changelog.

hg-fast-export.py
```diff
@@ -449,7 +453,7 @@ def verify_heads(ui,repo,cache,force,branchesmap):
 
   # verify that branch has exactly one head
   t={}
-  for h in repo.heads():
+  for h in repo.filtered('visible').heads():
```
frej: Will this and the `hidden()` checks above also filter out closed branches which has previously required using the `--force` option? (not a critique of this patch, just curious)

hg2git.py
```diff
@@ -35,7 +35,9 @@ def setup_repo(url):
   except TypeError:
     myui=ui.ui()
     myui.setconfig('ui', 'interactive', 'off')
-  return myui,hg.repository(myui,url)
+    # Avoids a warning when the repository has obsolete markers
+    myui.setconfig('experimental', 'evolution.createmarkers', True)
+  return myui,hg.repository(myui,url).unfiltered()
```
frej: Do this in a separate commit as it is not strictly required for the main functionality.

frej:
> It looks like an alternative here would be to avoid filtering over the numeric changeset ids and instead just filter over the filtered repository changelog.

Unless you think that would make a cleaner or smaller patch I think this variant is good enough considering that we use the numeric changeset ids both as markers to git fast-import and for remembering what has already been done.

hg-fast-export.py
Mossop: If I'm reading the mercurial code right then no, but revisions do have a closesbranch() property so it should be straightforward to detect those from the heads.

hg2git.py
Mossop: I don't use pull requests much, is the best way to do this to just close this request and open another two, one for each part?

frej:
> I don't use pull requests much, is the best way to do this to just
> close this request and open another two, one for each part?

Keep both commits in this pull reqest as they are related. Just split
the current commit into two and then force-push your "new" branch to
your "obsolete" branch on your github. Github will interpret that as you
updating the pull request.

Mossop:
Updated.

frej:
Looks good. Thank you for your contribution!

PR #180, opened 2019-11-22
Support submodules that are already git repositories.

Mossop:
Although disabled by default hg supports submodules that are either git
or svn repositories. In the git case at least there is no kind of
mapping necessary, we just need to write out the submodule reference
directory.

In this change we always call refresh_gitmodules regardless of whether
there were mappings specified. Git submodules are written out directly
and hg submodules are mapped if a mapping was included.

The changes here look larger than they actually are, none of the logic for finding and applying the mapping for the hg subrepo case is changed, it's just all indented one extra level.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -171,7 +171,8 @@ def refresh_gitmodules(ctx):
   gitmodules=""
   # Create the .gitmodules file and all submodules
   for name,subrepo_info in ctx.substate.items():
-    gitmodules+=refresh_hg_submodule(name,subrepo_info)
+    if name in submodule_mappings:
+      gitmodules+=refresh_hg_submodule(name,subrepo_info)
 
   if len(gitmodules):
     wr('M 100644 inline .gitmodules')
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -141,6 +141,13 @@ def remove_gitmodules(ctx):
       wr('D %s' % submodule)
   wr('D .gitmodules')
 
+def refresh_git_submodule(name,subrepo_info):
+  wr('M 160000 %s %s' % (subrepo_info[1],name))
+  sys.stderr.write("Adding/updating submodule %s, revision %s\n"
+                   % (name,subrepo_info[1]))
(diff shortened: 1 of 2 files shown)


frej:
This looks like a useful feature which I would like to merge!

> The changes here look larger than they actually are, none of the logic for finding and applying the mapping for the hg subrepo case is changed, it's just all indented one extra level.

Could you do the change in two commits? The first commit would move the body of the for loop  into its own function, making that step trivial to review. The second commit would then introduce the new functionality in its own function together with the logic picking the function to call. Hopefully that will make make my feeling for this patch go from "this could be right" to "this is obviously correct"

Mossop:
Updated

hg-fast-export.py
```diff
@@ -171,7 +178,10 @@ def refresh_gitmodules(ctx):
   gitmodules=""
   # Create the .gitmodules file and all submodules
   for name,subrepo_info in ctx.substate.items():
-    gitmodules+=refresh_hg_submodule(name,subrepo_info)
+    if subrepo_info[2]=='git':
+      gitmodules+=refresh_git_submodule(name,subrepo_info)
+    elif submodule_mappings and name in submodule_mappings:
```
frej: The `and name in submodule_mappings` should go into it's own patch as it does more than preserving the current functionality. As far as I can tell it is a new optimization which became simple to implement following the previous refactoring.

frej:
The code looks fine, I just want the commit messages to match what's done to the code, good work!

Mossop:
Sorry for the delay, been swamped with other stuff.

hg-fast-export.py
frej: Can you comment on the previous review comment as you didn't make a change, am I wrong or did you just miss to adress it?

Mossop: My apologies, let me fix that up.

frej: I looked at acf93a80a9e1b07c9158220c22aef3fcb8a1dd11, now I get it, thanks!

frej:
Now it looks good, thank you for your contribution.

Issue #182, opened 2019-12-17
Error on mercurial repo with both master and default branch


acrucifix:
Hello,

I am using this tool in order to migrate a bunch of repositories from mercurial to git. First of all, thank you very much for it, it helps me a lot in this migration.

Unfortunately, I stumbled upon a particular problem. In most of the repositories, the "default" branch initially generated by mercurial has been closed and replaced by a self-created "master" branch.

I have replicated this setup on the following public repo [https://bitbucket.org/acrucifix/test-hg2git/commits/](https://bitbucket.org/acrucifix/test-hg2git/commits/)

Using fast-export (commit `ffdd27c`) to migrate this repo to git, I have the following error:
```
Loaded 0 branches
Error: repository has at least one unnamed head: hg r8
/usr/local/Cellar/git/2.24.0_2/libexec/git-core/git-fast-import statistics:
---------------------------------------------------------------------
Alloc'd objects:       5000
Total objects:            0 (         0 duplicates                  )
      blobs  :            0 (         0 duplicates          0 deltas of          0 attempts)
      trees  :            0 (         0 duplicates          0 deltas of          0 attempts)
      commits:            0 (         0 duplicates          0 deltas of          0 attempts)
      tags   :            0 (         0 duplicates          0 deltas of          0 attempts)
Total branches:           0 (         0 loads     )
      marks:           1024 (         0 unique    )
      atoms:              0
Memory total:          2360 KiB
       pools:          2048 KiB
     objects:           312 KiB
---------------------------------------------------------------------
pack_report: getpagesize()            =       4096
pack_report: core.packedGitWindowSize = 1073741824
pack_report: core.packedGitLimit      = 35184372088832
pack_report: pack_used_ctr            =          0
pack_report: pack_mmap_calls          =          0
pack_report: pack_open_windows        =          0 /          0
pack_report: pack_mapped              =          0 /          0
---------------------------------------------------------------------
```

I tried reopening the default branch but it doesn't help.

Do you have any idea how we may solve this in order to keep the commits from both the default and master branches ?

Have a nice day,
Best,

frej:
[...]

PR #183, opened 2019-12-18
Document default branch behaviour

murrayju:
Document the behavior of renaming the default branch to master, and how to override when this causes problems. See #182.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,11 @@ branch/tag names. In the future -n will become the default, but in
 order to not break existing incremental conversions, the default
 remains with the old behavior.
 
+By default, the `default` mercurial branch is renamed to the `master` 
+branch on git. If your mercurial repo contains both `default` and 
+`master` branches, you'll need to override this behavior. Use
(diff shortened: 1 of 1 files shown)


frej:
I like the patch, but please shorten the subject to be less than 80 characters, [this recommends 50 characters, but I myself break that from time to time](https://chris.beams.io/posts/git-commit/l)

murrayju:
@frej I assume you are referring to the git commit message, but the first line of my commit (what I think you are calling the "subject") is only 16 characters.

Are you asking me to make it more descriptive, but limit that to 80 characters?

frej:
Sorry, I got confused by just having a quick look at the commit message on github and not with git log. What I should have said was (If I wasn't in a hurry): Make the subject more descriptive and make sure to line break the rest of the message to at most 80 chars.

frej:
Why not "Document default branch behaviour" as the subject? That it goes into the README is pretty obvious from the diffstat. The body of the commit message is not line folded to less than 80 characters.

frej:
Looks good now, thank you for your contribution!

PR #185, opened 2019-12-26
Made export_commit more reliable for different python/mercurial versions

kovasap:
This changed allowed me to run this tool on my linux machine.  It seems like it might be a more reliable way for others to run the tool as well.


frej:
Please have a look at [https://chris.beams.io/posts/git-commit/](https://chris.beams.io/posts/git-commit/) on how to write a good commit message, I'm particularly missing the *what* and *why* parts, in this case which Python/Mercurial versions do work, do not work etc. Is this patch triggered by an attempt to run Mercurial >= 5.2 on Python >= 3? If so please coordinate with and/or extend the work done in #167.

frej:
Superseded by #197 which provided the asked for information.

PR #186, opened 2019-12-31
Allow external URLs to be specified for submodule conversions

ccaughie:
This change allows remote URLs (e.g. Github URLs) to be written to the .gitmodules file instead of just local paths. This is important for allowing users to clone a converted repository with submodules without having to first clone all of the submodules to their machine first.

### DIFF
--- a/README-SUBMODULES.md
+++ b/README-SUBMODULES.md
@@ -18,9 +18,11 @@ subrepositories have been converted, a mapping file that maps the mercurial
 subrepository path to a converted git submodule path must be created. The
 format for this file is:
 
-"<mercurial subrepo path>"="<git submodule path>"
-"<mercurial subrepo path2>"="<git submodule path2>"
+```
+"<mercurial subrepo path>"="<git submodule path>[|<clone url>]"
+"<mercurial subrepo path2>"="<git submodule path2>[|<clone url2>]"
 ...
+```
 
 The path of this mapping file is then provided with the --subrepo-map
 command line option.
@@ -48,12 +50,25 @@ Create an empty new folder where all the converted git modules will be imported:
     git init
     hg-fast-export.sh -r ~/mercurial/subrepo2
 
+### Determine the remote URLs from which the submodules will be fetched
+
+These are the URLs that will be written to the .gitmodules file. Adding them to
+the mapping file will allow other users to update their submodules without first
+having to clone them to specific locations on their own machines.
+
+For this example we'll assume that the subrepos are going to be hosted on github:
+
+    submodule1 -> https://github.com/myname/submodule1.git
+    submodule2 -> https://github.com/myname/submodule2.git
+
 ### Create mapping file
     cd ~/imported-gits
     cat > submodule-mappings << EOF
-    "subrepo/subrepo1"="../submodule1"
-    "subrepo/subrepo2"="../submodule2"
+    "subrepo/subrepo1"="../submodule1|https://github.com/myname/submodule1.git"
+    "subrepo/subrepo2"="../submodule2|https://github.com/myname/submodule2.git"
     EOF
+
+Note that if the external URL is omitted, the local paths will be used instead.
 
 ### Convert main repository
     cd ~/imported-gits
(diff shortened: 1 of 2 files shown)


frej:
Please explain what problem you are trying to solve in the commit message, have a look at [https://chris.beams.io/posts/git-commit/](https://chris.beams.io/posts/git-commit/) for tips on on how to write good commit messages.

In the pull request you say: "This is important for allowing users to clone a converted repository with submodules without having to first clone all of the submodules to their machine first.". If you clone a git repo and then do `git submodule update --init` you will get the all the objects in the subrepos fetched into your parent's `.git` and all submodules checked out.

In the patch to `README-SUBMODULES.md` you say: "Adding them to the mapping file will allow other users to update their submodules without first having to clone them to specific locations on their own machines.". If you run `git submodule update` after a pull in the parent (bringing in changes to the subrepos), you will get the updates so I don't buy this motivation.

If by *updating* you mean doing incremental conversions, this patch will not avoid the requirement of having a full clones of the source Mercurial and the converted destination Git repos making up the parent and sub repos. You will need the `.git/hg2git-mapping` and `.git/hg2git-marks` files in the converted sub repos neither of which are available remotely.

I'm mainly confused with this patch, it claims to be motivated by two different needs, which are already provided with the unmodified master branch. Please clarify what problem you are trying to solve!

ccaughie:
Hi Frej,

The issue I had without this change (and maybe I missed something in how to use the tool) was that after conversion my .gitmodules file looked like this:

```
[submodule "mymodule"]
    path = mymodule
    url = ../mymodule
```

when I wanted it to look like this:

```
[submodule "mymodule"]
    path = mymodule
    url = https://our.git.server/mymodule
```

With the first version, `git submodule update` would only work if "mymodule" was already cloned into a sibling folder of the parent project (../mymodule). With the second version you can just clone the parent and then run `git submodule update --init` without having to manually clone the submodules first. This will automatically clone the submodules directly from the remote to their correct locations in the working dir of the parent project.

I couldn't just put the remote URL in the mapping file without also adding the local repo path however, because as you say the `.git/.hg2git-*` files are required for the conversion process.
[This blog](https://lyubomyr-shaydariv.github.io/posts/2019-02-06-hg-subrepos-to-git-submodules/) suggests another way of solving the problem using `git filter-branch` to post-process the converted repository. This would work for a single level of submodules, but would fail if you have submodules which themselves contain submodules (which we do), since the git revision IDs of the mid-level modules all change after running filter-branch, invalidating the mappings in the `.hg2git` files.

Point taken on the commit message, I'm not sure if there's a way to change it after the fact?

ccaughie:
Hmm I just re-read the git submodule documentation - it looks like it does resolve relative paths in .gitmodules by applying them to the default remote repository. So as long as I organized my local repos to mirror the hierarchy on our server I suppose could have left them as relative paths.

I still think this change adds a useful option though; I can think of a couple of reasons you might want to use absolute URLs in .gitmodules instead of relative ones:

1. You have a submodule that is on a different server than the parent (e.g. the parent is proprietary and on a corporate server, the submodule is open source and on github).
2. You want people to be able to fork the parent to their own server, or folder on the main server, without having to fork all of the submodules as well.

frej:
> Point taken on the commit message, I'm not sure if there's a way to change it after the fact?

[The Pro Git book has a chapter on rewriting history](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

> You have a submodule that is on a different server than the parent (e.g. the parent is proprietary and on a corporate server, the submodule is open source and on github).

If those external repos are controlled by you and incrementally converted by you, you'll still have to have a local clone, so I don't think the extra complexity is warranted just to avoid having a local clone on your corporate server. If the subrepos already exist on Github (and are referenced as native Git subrepos, then it just works as is, the support was added by #180).

> You want people to be able to fork the parent to their own server, or folder on the main server, without having to fork all of the submodules as well.

If you fork, nothing stops you from editing `.gitmodules` following the fork to add absolute paths (or filter branch to change it everywhere). This is not really relevant to fast export as this is no different from publishing a fork of a Git repo using submodules.

All in all, I feel that this PR adds complexity without a realistic use case and will therefore reject it.

ccaughie:
Totally up to you. Now that our Mercurial repos are all converted (with absolute submodule URLs throughout their history, which I still personally prefer) I doubt I'll ever need to use this again so I have no particular interest in my changes being merged into master.

Having said that, this tool saved me a ton of time so thank you for maintaining it. I'll leave my fork where it is in case others request this feature and you change your mind.

Issue #87, opened 2016-12-30
Conversion is incorrect due to the default core.ignorecase


emilianbold:
I have some changesets that just change the case of a file name (eg. from HTMLLexer.java to HtmlLexer.java).

hg-fast-export seems to produce an invalid git commit, because I am running conversion on macOS, where git init creates a repository with core.ignorecase=true.

I don't believe there is any way to fix this other than re-running the conversion (is it?).

If I run `git config core.ignorecase false` before executing hg-fast-export the conversion is correct.

How to reproduce the issue:

```sh
$ mkdir case-bug
$ cd case-bug/
$ mkdir m
$ cd m/
$ hg init .
$ echo 'a' > A
$ hg add A 
$ hg commit -m "Added A"
$ mv A a
$ ls
a
$ hg mv A a
$ hg st
A a
R A
$ hg commit -m 'Renaming A -> a'
$ hg st
$ hg log -p 
changeset:   1:b68723268fab
tag:         tip
user:        Emilian Bold <emilian.bold at gmail.com>
date:        Fri Dec 30 14:41:40 2016 +0200
summary:     Renaming A -> a

diff -r 5cffc220b750 -r b68723268fab A
--- a/A	Fri Dec 30 14:41:26 2016 +0200
+++ /dev/null	Thu Jan 01 00:00:00 1970 +0000
@@ -1,1 +0,0 @@
-a
diff -r 5cffc220b750 -r b68723268fab a
--- /dev/null	Thu Jan 01 00:00:00 1970 +0000
+++ b/a	Fri Dec 30 14:41:40 2016 +0200
@@ -0,0 +1,1 @@
+a

changeset:   0:5cffc220b750
user:        Emilian Bold <emilian.bold at gmail.com>
date:        Fri Dec 30 14:41:26 2016 +0200
summary:     Added A

diff -r 000000000000 -r 5cffc220b750 A
--- /dev/null	Thu Jan 01 00:00:00 1970 +0000
+++ b/A	Fri Dec 30 14:41:26 2016 +0200
@@ -0,0 +1,1 @@
+a

$ cd ..
$ mkdir g
$ cd g/
$ git init .
Initialized empty Git repository in /Users/apache/work/case-bug/g/.git/
$ ~/work/fast-export/hg-fast-export.sh -r ../m/
master: Exporting full revision 1/2 with 1/0/0 added/changed/removed files
master: Exporting simple delta revision 2/2 with 1/0/1 added/changed/removed files
[...]

Issue #188, opened 2020-01-16
Same-named local subrepos with different origins cannot be converted due to ambiguity in the mapping


jdanielp:
Please see attached an example repo and mapping file that cannot be converted with the given submodule mapping technique.  

Top level repo is srt_ambiguous.  In v1 and v2 branches there is a srt_subx subrepo.  But, they have different origins per branch.  in branch v1 srt_subx corresponds to srt_suba repo.  In v2 branch srt_subx refers to srt_subb repo.

I suspect the mapping file convention will need to be extended to define per-branch mappings.  As with issue 187, I am not familiar-enough with the inner workings of hg2git and hg-fast-export(.sh|.py) to hack this myself.

[ambiguous.zip](https://github.com/frej/fast-export/files/4073578/ambiguous.zip)

frej:
When extending the mapping file format, be sure to also handle the case when a subrepo is changed without a change in the current branch. I'm thinking of a case when, for example, you have been using an internally generated conversion of a Subversion repo, but when upstream switched to Mercurial, you started using that repo directly.

jdanielp:
I'm making my way through this.  It's doable without too much surgery so far as I can tell.  I have `submodule_mappings` converted to deal with different `url`s in different branches. Now I need to deal with the `mapping_cache`, `marks_cache`, and `subrepo_cache`.  Those give me some concern since I'll need to learn their usage, etc.

@frej Do you have an example repo for the scenario you describe?

frej:
@jdanielp: Sorry, no I don't have an example. Another thing, make sure that you stay backwards compatible so that existing incremental setups are not broken.

jdanielp:
@frej 

I was able to get a solution in place this morning.  That solution works for the example repository attached on this issue.

Then I realized that I had to get my validation script working so I could validate bigger repos automatically rather than manually spot checking a few commits.  I have that script working now (hopefully), but it's late, so I will verify my changes work Wednesday.

The basic idea of my changes is that I allow the user to specify a json file for `--subrepo-map`.  No .sh script changes needed.  The code to load the mapping sees that it's a .json and uses a different function I added.

With json, the user can define a per-branch mapping in a file format that lends itself to nested dicts.  For example, here is the mapping file for the srt_ambiguous example repo:
```
{
    "srt_subx": { "": "c:/Users/Jpietkiewicz/repos/conversion/git/srt_suba",
[...]

PR #189, opened 2020-01-22
Refresh submodule list when .hgsub changes

### LINKED ISSUE #187: Subrepos that have been removed are not converted correctly, cannot init submodules
Please see attached zip of an example mercurial repo with subrepos wherein one of the subrepos is not used in a certain branch after a certain change set.  

srt stands for "SubRepo Test".  The top level repo is srt_top, and it has two subrepos srt_suba and srt_subb.  Sorry for the naming; I apparently did not eat my Wheaties this morning. :)

For some time both subrepos are used by the top-level repo, but at change set e88d9415ea8e1a2a025b9df4c1054f2c2787cc72 srt_suba is removed from use by removing its entry in .hgsub.

I cloned out the subrepos to their own folders outsidee srt_top.  I converted the subrepos first, then the top level via the steps in README-Submodules.md

When I do the following in the git repo
`git checkout v3`
`git submodule update --init`

I receive the error:
> fatal: No url found for submodule path 'srt_suba' in .gitmodules

This also happens when using [Sourcetree](https://www.sourcetreeapp.com/) to navigate between branches.  This is the full output when attempting to checkout v3 branch:

> git -c diff.mnemonicprefix=false -c core.quotepath=false --no-optional-locks checkout v3 --progress
> M	srt_subb
> Switched to branch 'v3'
> git -c  diff.mnemonicprefix=false -c core.quotepath=false --no-optional-locks submodule update --init
> fatal: No url found for submodule path 'srt_suba' in .gitmodules
> Completed with errors, see above.

I have found some forum posts like [this](https://community.atlassian.com/t5/Sourcetree-questions/SourceTree-says-No-url-found-for-submodule-path/qaq-p/628869) and [this](https://stackoverflow.com/questions/4185365/no-submodule-mapping-found-in-gitmodule-for-a-path-thats-not-a-submodule) and doing the "git rm --cached X" will work, but only forward in time.  I tried re-writing history to add a commit that does the "git rm...." with "git rebase -i" via [this post](https://stackoverflow.com/questions/32315156/how-to-inject-a-commit-between-some-two-arbitrary-commits-in-the-past) but the interactive rebase had some merge problems in my actual repo and I'd rather not have to re-write history every changeset that removes a subrepo.

Please let me know if there is anything you need from me.  I am in charge of converting our mercurial repos to git repos because [bitbucket](https://bitbucket.org/blog/sunsetting-mercurial-support-in-bitbucket).  Right now I'm not familiar enough with the hg-fast-export tool to make the change myself and submit a pull request.  I am happy to beta test, though I suspect the attached example will be sufficient for verifying the first draft.

[srt_top.zip](https://github.com/frej/fast-export/files/4060496/srt_top.zip)
Edit: Adding 
[srt_top_submodule_mappings.txt](https://github.com/frej/fast-export/files/4073560/srt_top_submodule_mappings.txt)

MokhamedDakhraui:
Refresh list of submodules when `.hgsub` file changes.

close #187

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -191,8 +191,10 @@ def refresh_gitmodules(ctx):
 def export_file_contents(ctx,manifest,files,hgtags,encoding='',plugins={}):
   count=0
   max=len(files)
+  is_submodules_refreshed=False
   for file in files:
-    if file==".hgsubstate":
+    if not is_submodules_refreshed and (file=='.hgsub' or file=='.hgsubstate'):
+      is_submodules_refreshed=True
       refresh_gitmodules(ctx)
     # Skip .hgtags files. They only get us in trouble.
     if not hgtags and file == ".hgtags":
@@ -336,7 +338,7 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
     if fn_encoding:
       filename=filename.decode(fn_encoding).encode('utf8')
     filename=strip_leading_slash(filename)
-    if filename=='.hgsubstate':
+    if filename=='.hgsub':
       remove_gitmodules(ctx)
     wr('D %s' % filename)
 


MokhamedDakhraui:
This is sloppy hotfix for now. At least it allows to set label `patch-exists` :smile:

hg-fast-export.py
```diff
@@ -336,7 +336,7 @@ def get_branchname(name):
     if fn_encoding:
       filename=filename.decode(fn_encoding).encode('utf8')
     filename=strip_leading_slash(filename)
-    if filename=='.hgsubstate':
+    if filename=='.hgsubstate' or filename=='.hgsub':
```
MokhamedDakhraui: Use `context.substate` to check subrepos

Smth like:
```suggestion
    if (filename=='.hgsubstate' or filename=='.hgsub') and not context.substate:
```

jdanielp:
@MokhamedDakhraui I just tested this change with the srt_top repo attached to #187.  It does not fix the issue.  This change only happens when the `.hgsub` or `.hgsubstate` file is removed, but your comment above says it should happen when the file is changed.  There is no such loop for changed files in the `export_commit` function, only for removed files.

MokhamedDakhraui:
Before make PR I tested it on `srt_top` repo from #187. On v3 command git submodules shows only `srt_subb`.
Submodules list should be refreshed when .hgsub-ish files added/modified here:
https://github.com/frej/fast-export/blob/4e8c6bbc9747bffc6abbb62b7f6a725d82989b44/hg-fast-export.py#L195

Can you apply resulted git repo and point to commit after which problem occurs.
I also replay test on my end latter

MokhamedDakhraui:
I'm just contributor so I the final decision will make @frej 

If just for opinion, then it seems that your patch do the same as `refresh_submodules` but more selectively (not just remove all submodules and then readd relevant). So I suggest you to refactor `refresh_submodules` to add/remove submodules in more optimal way in new PR

jdanielp:
@MokhamedDakhraui I tested your fix on my work repo and it looks good.  I am able to checkout the various branches without issue.  

For the record: I did not include the suggested change for `and not context.substate` on line 339

alexey-lcard:
Fix works great on my repos too. Many thanks!

hg-fast-export.py
```diff
@@ -192,7 +192,7 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding='',plugins={}):
   count=0
   max=len(files)
   for file in files:
-    if file==".hgsubstate":
+    if file=='.hgsub' or file='.hgsubstate':
```
frej:
```Python
file='.hgsubstate'
```
Isn't valid python. (`==` is probably what you intended, and with that this patch seems to fix #87)

frej:
@MokhamedDakhraui : I really appreciate that you support your previous contributions. I'm much more likely to accept patches when I know that the author will provide support if bugs turn up later on. Thanks for the help!

hg-fast-export.py
MokhamedDakhraui: Fixed

MokhamedDakhraui:
Optimized a little to refresh submodules only once after crossing over either `.hgsub` or `.hgsubstate`. I guess it's ready for review

jdanielp:
@frej I'm fine with what you said about using @MokhamedDakhraui 's fix.  If you want me to test the fix and review the changes as a second reviewer I can do that on Monday.

I submited the patch as a zip since I cannot submit .diff or .patch files as attachments.  Should I have initiated my own pull request with my patch?  I did not do so because it would have split the conversation onto two different threads if I submitted as a new PR.  Or maybe that's what's supposed to happen.  I'm a newbie when it comes to working with open source projects.

I plan to work on #188 early next week.  If I get to a fix first I'll be sure to use a PR there.

frej:
@jdanielp, I'll take you up on the offer to run the patch on your internal repo, let me know how it goes.

jdanielp:
@frej @MokhamedDakhraui @alexey-lcard 

I applied the patch from https://github.com/frej/fast-export/pull/189.patch to my working copy and converted my work repo.  Everything looks good when I switch between branches in sourcetree and in git bash on windows.  I'd say we're good to go.

PR #193, opened 2020-01-29
Adding a new plugin to overwrite null messages.

phargogh:
This PR adds a plugin to overwrite null commit messages (which git doesn't handle very well) with a user-defined string.  A reasonable default is assumed.

The use case for this is that bitbucket allows you to create a mercurial snippet with an empty commit message.  When the repository is converted with `hg-fast-export`, I'm then unable to push the repository to github because of this null message.


frej:
LGTM, will merge. Thank you for your contribution.

PR #194, opened 2020-01-29
Handle `--force` option correctly in any position

### LINKED ISSUE #57: hg-fast-export bash script breaks when --force is used
When invoking:
  hg-fast-export --force --quiet -r /path/to/hg/repo

The script bails out saying one should provide a '-r' option, which is provided.

I suspect the "break" statement in the bash loop collecting arguments is unintentional.

### LINKED ISSUE #190: `--force` option is not accepted
If I specify it as the first option, I get:

```
no repo given, use -r flag
```

If I specify it as any subsequent option (including the last), I get:

```
Error: The option core.ignoreCase is set to true in the git
repository. This will produce empty changesets for renames that just
change the case of the file name.
Use --force to skip this check or change the option with
git config core.ignoreCase false
```

alexreg:
Fixes #57, fixes #190.

### DIFF
--- a/hg-fast-export.sh
+++ b/hg-fast-export.sh
@@ -110,7 +110,7 @@ if [ "true" = "$IGNORECASE" ]; then
     IGNORECASEWARN="true"
 fi;
 
-
+HG2GIT_ARGS=()
 while case "$#" in 0) break ;; esac
 do
   case "$1" in
@@ -125,14 +125,10 @@ do
       # pass --force to git-fast-import and hg-fast-export.py
       GFI_OPTS="$GFI_OPTS --force"
       IGNORECASEWARN="";
-      break
       ;;
-    -*)
-      # pass any other options down to hg2git.py
-      break
-      ;;
-    *)
-      break
+    # pass all other args down to hg2git.py
+	*)
+      HG2GIT_ARGS+=("$1")
       ;;
   esac
   shift
@@ -187,7 +183,7 @@ $(
       --mapping "$GIT_DIR/$PFX-$SFX_MAPPING" \
       --heads "$GIT_DIR/$PFX-$SFX_HEADS" \
       --status "$GIT_DIR/$PFX-$SFX_STATE" \
-      "$@" 3>&- || _e1=$?
+      "${HG2GIT_ARGS[@]}" 3>&- || _e1=$?
     echo $_e1 >&3
   } | \
   {


hg-fast-export.sh
```diff
@@ -187,7 +183,7 @@ $(
       --mapping "$GIT_DIR/$PFX-$SFX_MAPPING" \
       --heads "$GIT_DIR/$PFX-$SFX_HEADS" \
       --status "$GIT_DIR/$PFX-$SFX_STATE" \
-      "$@" 3>&- || _e1=$?
+      s"${HG2GIT_ARGS[@]}" 3>&- || _e1=$?
```
frej: Does the  `s"` have a special meaning in sh or is something that got there by mistake? I think the latter as I can't get it to do something special on the command line.

alexreg: Oops, merely a typo, sorry.

alexreg:
No problem! Glad I could help in a little way.

frej:
And squash the commits please, there is no need to preserve the history of your typo :)

alexreg:
Agreed. Unfortunately GitHub doesn't have the feature (I was doing this in-browser), so let me clone and do that quickly...

frej:
That was fast, once again thank you for your contribution!

alexreg:
@frej Your intention was to support the old Bourne shell, I guess? Unfortunately arrays are only in Bash. This script becomes quite hard to write without arrays (I mean, my changes.)

frej:
The changes turned out to require bash. Traditionally we have tried to stay compatible with plain old sh, so this is a revert.

alexreg:
Sorry, I am so used to using Bash I forgot such a basic feature as arrays are not in the Bourne Shell... I wonder if quoting of arguments can be done correctly in that case.

alexreg:
@frej I suppose there's no way I can convince you into switching with Bash? I think it's pretty-much installed by default on all *nix systems (at least version 3).

frej:
Git, as far as I can tell, does not require bash, so I don't think it is
a good move to introduce that as requirement.

The last time I, by mistake, introduced a bashism, I seem to remember
getting a bug report along the lines of "on X there is no bash, just
Y. Please patch it like this and it will work without bash on my Z
system", so apparently there are systems out there which do not have
bash. Also, considering that your ticket for this gotcha is the first
one in five years, I don't think that this is a serious enough problem
to motivate adding a new requirement.

alexreg:
Fair enough.

PR #198, opened 2020-02-07
Fix issue #166: Ignore files and directories called .git

### LINKED ISSUE #166: Conversion gotcha if the source repo contains a file named .git
Hello, this is _not_ a urgent issue, more like a gotcha/caveat.

I have used fast-export to convert a repo of a project containing a tiny plain text file named `.git`, which was part of a small library whose sources had been included in a subdirectory of the source tree. Inside the file was a path that pointed to a directory two levels up, but the actual contents didn't seem to matter.

When the conversion was done, this happened (after running `fast-export` on a repro repo with just a .git text file committed): 

    $ git status

    On branch master
    Changes to be committed:
      (use "git restore --staged <file>..." to unstage)
            deleted:    .git

    $ git checkout HEAD

    error: Updating the following directories would lose untracked files in them:
            .git

    Aborting

    $ git reset --hard HEAD
    error: invalid path '.git'
    HEAD is now at 2a047b8 create .git file

I solved the problem easily by simply getting rid of the file which had no business being there in the first place and ran a new conversion.

So it seems that one of the tools in the chain does not like files named `.git`.

Not surprising, since Mercurial itself will refuse to initialize a repository in a directory that contains such a file, saying `abort: Cannot create a git repository.`

Again: not a problem that needs urgent attention, maybe just a small note on the README would do. I can see how it could potentially impact other people and cause some head scratching.

Thanks for this wonderful tool, I've tested it successfully with several different repos and this is the only glitch I noticed.

chrisjbillington:
Ignore files and directories called `.git`
Git cannot track these files. Print a warning if encountering one.

Fixes #166

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,11 @@ hg-fast-export supports multiple branches but only named branches with
 exactly one head each. Otherwise commits to the tip of these heads
 within the branch will get flattened into merge commits.
 
+hg-fast-export will ignore any files or directories tracked by mercurial
+called `.git`, and will print a warning if it encounters one. Git cannot
+track such files or directories. This is not to be confused with submodules,
+which are described in README-SUBMODULES.md.
+
 As each git-fast-import run creates a new pack file, it may be
 required to repack the repository quite often for incremental imports
 (especially when importing a small number of changesets per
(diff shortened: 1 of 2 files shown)


frej:

chrisjbillington:
No worries, thanks for merging :)

PR #199, opened 2020-02-07
Add additional options to branch_name_in_commit plugin

chrisjbillington:
- Allow skipping writing the branch name if the branch is 'master'.

- Allow writing the branch name on the same line as the first line of
  the commit message separated by a colon, instead of it having its own
  line.

This is backward compatible, the previous argument format is still valid.


frej:
LGTM, will merge. Thank you for your contribution.

chrisjbillington:
Thanks!

PR #201, opened 2020-02-11
Support Python 3

chrisjbillington:
Port hg-fast-import to Python 2/3 polyglot code.

Since mercurial accepts and returns bytestrings for all repository data,
the approach I've taken here is to use bytestrings throughout the
hg-fast-import code. All strings pertaining to repository data are
bytestrings. This means the code is using the same string datatype for
this data on Python 3 as it did (and still does) on Python 2.

Repository data coming from subprocess calls to git, or read from files,
is also left as the bytestrings either returned from
subprocess.check_output or as read from the file in 'rb' mode.

Regexes and string literals that are used with repository data have
all had a b'' prefix added.

When repository data is used in error/warning messages, it is decoded
with the UTF8 codec for printing.

With this patch, hg-fast-export.py writes binary output to
sys.stdout.buffer on Python 3 - on Python 2 this doesn't exist and it
still uses sys.stdout.

The only strings that are left as "native" strings and not coerced to
bytestrings are filepaths passed in on the command line, and dictionary
keys for internal data structures used by hg-fast-import.py, that do
not originate in repository data.

Mapping files are read in 'rb' mode, and thus bytestrings are read from
them. When an encoding is given, their contents are decoded with that
encoding, but then immediately encoded again with UTF8 and they are
returned as the resulting bytestrings

Other necessary changes were:

 - indexing byestrings with a single index returns an integer on Python 3.
   These indexing operations have been replaced with a one-element
   slice: x[0] -> x[0:1] or x[-1] -> [-1:] so as to return a bytestring.

 - raw_hash.encode('hex_codec') replaced with binascii.hexlify(raw_hash)

 - str(integer) -> b'%d' % integer

 - 'string_escape' codec replaced with 'unicode_escape' (which was
    backported to python 2.7). Strings decoded with this codec were then
    immediately re-encoded with UTF8.

 - Calls to map() intended to execute their contents immediately were
   unwrapped or converted to list comprehensions, since map() is an
   iterator and does not execute until iterated over.

hg-fast-export.sh has been modified to not require Python 2. Instead, if
PYTHON has not been defined, it checks python2, python, then python3,
and uses the first one that exists and can import the mercurial module.

I've tested that this patch produces identical output from hg-fast-export.py when run with Python 2 or 3 on an example repo with a few branches, mercurial subrepos, an author mapping file, and every plugin used.

Running the test with coverage.py gives the following coverage:
```
file                                            hit     missed  skipped  %
---------------------------------------------------------------------------
hg-fast-export.py                               467     91      0       81%
hg2git.py                                       100     14      0       86%
pluginloader/__init__.py                        17      3       0       82%
plugins/branch_name_in_commit/__init__.py       20      3       0       85%
plugins/dos2unix/__init__.py                    9       0       0       100%
plugins/issue_prefix/__init__.py                10      0       0       100%
plugins/shell_filter_file_contents/__init__.py  24      4       0       83%
```

I don't really know what hg-reset.py is for, I've ported it too, and whilst it's pretty simple code, that is untested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,10 @@ first time.
 System Requirements
 -------------------
 
-This project depends on Python 2.7 and the Mercurial >= 4.6
-package. If Python is not installed, install it before proceeding. The
-Mercurial package can be installed with `pip install mercurial`.
+This project depends on Python 2.7 or 3.5+, and the Mercurial >= 4.6
+package (>= 5.2, if Python 3.5+). If Python is not installed, install
+it before proceeding. TheMercurial package can be installed with
+`pip install mercurial`.
 
 On windows the bash that comes with "Git for Windows" is known to work
 well.
--- a/hg-fast-export.sh
+++ b/hg-fast-export.sh
@@ -28,25 +28,20 @@ SFX_STATE="state"
 GFI_OPTS=""
 
 if [ -z "${PYTHON}" ]; then
-    # $PYTHON is not set, so we try to find a working python 2.7 to
-    # use. PEP 394 tells us to use 'python2', otherwise try plain
-    # 'python'.
-    if command -v python2 > /dev/null; then
-	PYTHON="python2"
-    elif command -v python > /dev/null; then
-	PYTHON="python"
-    else
-        echo "Could not find any python interpreter, please use the 'PYTHON'" \
-	     "environment variable to specify the interpreter to use."
-        exit 1
-    fi
+    # $PYTHON is not set, so we try to find a working python with mercurial:
+    for python_cmd in python2 python python3; do
+        if command -v $python_cmd > /dev/null; then
+            $python_cmd -c 'import mercurial' 2> /dev/null
+            if [ $? -eq 0 ]; then
+                PYTHON=$python_cmd
+                break
+            fi
+        fi
+    done
 fi
-
-# Check that the python specified by the user or autodetected above is
-# >= 2.7 and < 3.
-if ! ${PYTHON} -c 'import sys; v=sys.version_info; exit(0 if v.major == 2 and v.minor >= 7 else 1)' > /dev/null 2>&1 ; then
-    echo "${PYTHON} is not a working python 2.7 interpreter, please use the" \
-	 "'PYTHON' environment variable to specify the interpreter to use."
+if [ -z "${PYTHON}" ]; then
+    echo "Could not find a python interpreter with the mercurial module available. " \
+        "Please use the 'PYTHON' environment variable to specify the interpreter to use."
     exit 1
 fi
 
--- a/plugins/issue_prefix/__init__.py
+++ b/plugins/issue_prefix/__init__.py
@@ -7,9 +7,11 @@ def build_filter(args):
 
 class Filter:
     def __init__(self, args):
+        if not isinstance(args, bytes):
+            args = args.encode('utf8') 
         self.prefix = args
 
     def commit_message_filter(self, commit_data):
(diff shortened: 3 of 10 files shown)


chrisjbillington:
In case it isn't clear, this is a real port tested and working with actual mercurial on Python 3, as opposed to the other two PRs for Python 3 compat which were made without knowledge of what mercurial's behaviour on Python 3 turned out to be (i.e. a royal pain in the neck).

README.md
```diff
@@ -29,7 +29,7 @@ first time.
 System Requirements
 -------------------
 
-This project depends on Python 2.7 and the Mercurial >= 4.6
+This project depends on Python 2.7 or 3.5+, and the Mercurial >= 4.6
```
frej: Perhaps it should say: "... and the Mercurial package >= 4.6 (>= 5/2, if Python 3.5+)" to avoid future support issues.

hg-fast-export.sh
```diff
@@ -28,25 +28,20 @@ SFX_STATE="state"
...
-# Check that the python specified by the user or autodetected above is
-# >= 2.7 and < 3.
-if ! ${PYTHON} -c 'import sys; v=sys.version_info; exit(0 if v.major == 2 and v.minor >= 7 else 1)' > /dev/null 2>&1 ; then
-    echo "${PYTHON} is not a working python 2.7 interpreter, please use the" \
-	 "'PYTHON' environment variable to specify the interpreter to use."
+if [ -z "${PYTHON}" ]; then
+    echo "Could not find a python interpreter with the mercurial module available. " \
+        "Please use the 'PYTHON'environment variable to specify the interpreter to use."
     exit 1
 fi
```
frej: Neat, this is much better than what we had before,

plugins/issue_prefix/__init__.py
```diff
@@ -7,7 +7,9 @@ def build_filter(args):
 
 class Filter:
     def __init__(self, args):
-        self.prefix = args.encode('utf8')
+        if not isinstance(args, bytes):
+            args = args.encode('utf8') 
+        self.prefix = args
```
frej: Please squash this into the previous commit as it has not gotten into master yet and as I had a suggestion for the readme.

frej:
BTW @chrisjbillington, do you mind if I link to your hg-export-tool in the readme?

chrisjbillington:
Thanks for your kind words! I've modified the README regarding mercurial's version on Python 3, and squashed.

Yes, feel free to link to `hg-export-tool` :). Its functionality (basically, dealing with anonymous/bookmarked heads) should probably be absorbed into `hg-fast-export` at some point, but hopefully useful to point others to it in the meantime.

frej:

frej:
@chrisjbillington: I'm open to absorb the hg-export-tool into fast-export as long as we don't hard-code the naming of duplicate heads. We can provide a default but we should not force a particular policy.

chrisjbillington:
> @chrisjbillington: I'm open to absorb the hg-export-tool into fast-export as long as we don't hard-code the naming of duplicate heads. We can provide a default but we should not force a particular policy.

Yes, I agree - not sure how to best go about it, but something to think about!

At the moment how my tool is implemented, the branch name creation precedes any filters in `hg-fast-export`, so a branch name mapping file will still work with my tool - it will just have to have <branchname>-<n> (or the bookmark name) on the LHS.

So perhaps that will be sufficient, so long as whatever way it is absorbed also has the renaming preceding processing the mapping file, and so long as it is documented that you'll need to make your branch mapping file in that way, and that the head numbering is done in reverse timestamp order, so it is deterministic.

Issue #203, opened 2020-02-19
Line 359 breaks when branchname uses unicode


EraYaN:
OS: Ubuntu 19.10
Hg: 4.8.2
Python:  3.7.5
fast-export: 160aa3c

The error is:

```
Traceback (most recent call last):
  File "/fast-export/hg-fast-export.py", line 718, in <module>
    plugins=plugins_dict))
  File "/fast-export/hg-fast-export.py", line 570, in hg2git
    plugins)
  File "/fast-export/hg-fast-export.py", line 359, in export_commit
    (branch.decode('utf8'),type,revision+1,max,len(added),len(changed),len(removed)))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xeb' in position 5: ordinal not in range(128) 
```

In this case the branch name is `belgië` and the repo codec is cp1252. If I remove the `.decode('utf-8')` part from the branch variable, it works.
I'm not sure why that one was added, if it's not fully required, I can make a PR to remove it if you want.

chrisjbillington:
Interestingly the decode() of the bytestring is not what's failing, it's the encode() upon printing it to your terminal. This makes it seem like the issue is that your terminal doesnt support unicode. Weird for Ubuntu - what terminal are you using? What's your LANG environment variable?.

I'll see what I can do - since it's just for printing we could use any old error handler or just print the repr of the bytestring by removing the decode() as you've done, but it would be nice if it worked! Or we can just write the raw bytes using stderr.buffer and leave the terminal to deal with it. 

I would have expected an 'ë' to fail decoding UTF8 if it were actually cp1252 - could you perhaps print the repr() of that bytestring to stderr so I can see which encoding it is?

chrisjbillington:
For what it's worth I did test non ascii branch names, so they work in other cases, I think this must be specific to your environment.

Can you run a python script that just says print("belgië")? I'm kinda expecting that to raise the same error on your system.

EraYaN:
Alright so this is a fairly fresh Ubuntu install.

```
$ echo $LANG
en_US.UTF-8
```

and for the direct print:
```
>>> print("belgië")
belgië
```
my terminal is bash 5.0.3 in whatever the default (gnome?) emulator is.

Adding a repr print to the script it appears it's even weirder, no idea what the encoding is.
```
'belgi\xc3\xab'
```

chrisjbillington:
Cool, thanks. What about sys.stderr.write("belgië") from within hg-fast-export.py?

[...]

PR #204, opened 2020-02-19
Resolve issue #203: Encode stderr with UTF8 on PY2

chrisjbillington:
In Python 3, `sys.stderr.write()` requires unicode strings, and all
output on standard streams is UTF8 encoded. Therefore in the port to
Python 3, we `.decode()`d all strings that are used in `%` formatting of
strings to be printed to stderr.

However, in Python 2, `sys.stderr` accepts either bytestrings or unicode
strings, and:

- `%s` formatting of a bytestring with a unicode string, i.e  `"%s" %
  u"foo"` results in a unicode string.
- Writing a unicode string to stderr/stdout uses that stream's encoding
- When the output of the process is being piped somewhere other than a
  terminal (as it is when called with pipes and shell redirection from
  hg-fast-export.sh), that encoding is None, which implies ASCII.
- This raises UnicodeEncodeError if the unicode strings passed to
  `stderr.write()` have non-ascii characters.

Therefore we should not write unicode strings to stderr in Python 2. All
strings should be encoded UTF8 first.

This commit does that by using `codecs.getwriter("utf8")(sys.stderr)` as
the file-like object to write to instead of `sys.stderr` directly. This
brings the behaviour of writing to stderr on Python 2 in line with that
of Python 3.

The outputting to stderr in hg2git.py and
`shell_filter_file_contents/__init__.py` only use "native" strings and
so do not need to be modified to work correctly on both Python versions.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -40,6 +40,7 @@ submodule_mappings=None
 auto_sanitize = None
 
 stdout_buffer = sys.stdout if PY2 else sys.stdout.buffer
+stderr_buffer = sys.stderr if PY2 else sys.stderr.buffer
 
 def gitmode(flags):
   return b'l' in flags and b'120000' or b'x' in flags and b'100755' or b'100644'
@@ -57,7 +58,7 @@ def wr(msg=b''):
 def checkpoint(count):
   count=count+1
   if cfg_checkpoint_count>0 and count%cfg_checkpoint_count==0:
-    sys.stderr.write("Checkpoint after %d commits\n" % count)
+    stderr_buffer.write(b"Checkpoint after %d commits\n" % count)
     wr(b'checkpoint')
     wr()
   return count
@@ -150,9 +151,8 @@ def remove_gitmodules(ctx):
 
 def refresh_git_submodule(name,subrepo_info):
   wr(b'M 160000 %s %s' % (subrepo_info[1],name))
-  sys.stderr.write(
-    "Adding/updating submodule %s, revision %s\n"
-    % (name.decode('utf8'), subrepo_info[1].decode('utf8'))
+  stderr_buffer.write(
+    b"Adding/updating submodule %s, revision %s\n" % (name, subrepo_info[1])
   )
   return b'[submodule "%s"]\n\tpath = %s\n\turl = %s\n' % (name, name, subrepo_info[0])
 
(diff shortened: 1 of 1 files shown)


EraYaN:
This appears to fix the issue even if the reported encoding of stderr is `None` (when running as a child from a `subprocess.run` call.

chrisjbillington:
I'm actually not quite happy with this.

UTF8 is not quite ok, since filenames can actually be invalid UTF8 on unix. We would need to use the `surrogateescape` error handler on Py3 so that UTF8-invalid filepaths round-trip, and just not decode on Python2 (where that error handler doesn't exist).

I'll figure it out!

chrisjbillington:
Ok, I've resolved my further concerns about filenames potentially not even being valid UTF8.

@EraYaN could you test again?

New commit message below:

----

In Python 3, `sys.stderr.write()` requires unicode strings, and all
output on standard streams is UTF8 encoded. Therefore in the port to
Python 3, we `.decode()`d all strings that are used in `%` formatting of
strings to be printed to stderr.

However, in Python 2, `sys.stderr` accepts either bytestrings or unicode
strings, and:

- `%s` formatting of a bytestring with a unicode string, i.e  `"%s" %
  u"foo"` results in a unicode string.
- Writing a unicode string to stderr/stdout uses that stream's encoding
- When the output of the process is being piped somewhere other than a
  terminal (as it is when called with pipes and shell redirection from
  hg-fast-export.sh), that encoding is None, which implies ASCII.
- This raises UnicodeEncodeError if the unicode strings passed to
  `stderr.write()` have non-ascii characters.

We cannot fix this problem simply by encoding UTF8 again before writing
to stderr on Python 2. This is because the *decoding* of filenames with
the UTF8 codec may fail - filenames may not even be valid UTF8 desite
this being the declared filesystem encoding.

We could `fsdecode()` filenames on Python 3, which would use the
`surrogateescape` error handler, but stderr does not use this error
handler for output, meaning we would just have to encode again (with the
same error handler) anyway. And Python 2 lacks the `surrogateescape`
error handler in any case - we would need to reimplement it just to do a
round-trip decode and encode for no reason.

This commit leaves filenames and other repository data as bytestrings,
and simply writes them to `sys.stderr.buffer` on Python 3 or
`sys.stderr` on Python 2 as-is, after `%` formatting with bytestring
literals. This avoids encoding issues of filenames altogether.

Other writing to stderr that does not involve repository data has been
left with "native" strings, i.e.
`sys.stderr.write("a string literal %s" % a_command_line_arg)`. These
will still fail on Python 3 if the user passes a non-UTF filename as a
command line argument or similar. This is acceptable IMHO - although
`hg-fast-export` may encounter invalid UTF8 in mercurial repositories,
it is not too much to impose that the user name their branch mapping
files etc with valid UTF8!

EraYaN:
With the new changes, it still works.

frej:
Nicely done @chrisjbillington, I greatly appreciate you taking responsibility for, and fixing issues discovered in your previous work. Thank you for your contribution!

chrisjbillington:
You're welcome @frej! Thanks for merging.

Issue #207, opened 2020-02-29
When an unnamed head is found, fast-export crashes with a type error


madmaxreach:
Rain into this when converting a repo this weekend.

My environment is Mac OS X Catalina, with python and mercurial installed via homebrew.

While converting, I encountered the following error:

> Traceback (most recent call last):
  File "/Users/maxfuqua/temp/fast-export/hg-fast-export.py", line 711, in <module>
    plugins=plugins_dict))
  File "/Users/maxfuqua/temp/fast-export/hg-fast-export.py", line 527, in hg2git
    if not verify_heads(ui,repo,heads_cache,force,branchesmap):
  File "/Users/maxfuqua/temp/fast-export/hg-fast-export.py", line 497, in verify_heads
    % repo.changelog.rev(h)
TypeError: %b requires a bytes-like object, or an object that implements __bytes__, not 'int'

This appears to be related to PR #204 where output to stderr are now bytestrings.

PR #208, opened 2020-02-29
fix for crash specified in issue 207

madmaxreach:
Fixing crash with a type error.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -493,7 +493,7 @@ def verify_heads(ui,repo,cache,force,branchesmap):
     (_,_,_,_,_,_,branch,_)=get_changeset(ui,repo,h)
     if t.get(branch,False):
       stderr_buffer.write(
-        b'Error: repository has at least one unnamed head: hg r%s\n'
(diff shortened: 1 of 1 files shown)


frej:
@madmaxreach : Please include the reason for the change in the commit message (please see [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) for what constitutes a good commit message) instead of just referring to the issue number.

@chrisjbillington: You're my Python 3 expert, as far as I can tell this will work both for Python 2 and 3, correct?

chrisjbillington:
@frej yes, this looks like the correct fix to me.

Issue #207 is Python 3 specific, Python 2 doesn't mind if you give an int to `b'%s'`, but Python 3 refuses to automatically convert. Both are happy with `b'%d' % an_integer`.

madmaxreach:
@frej commit message should now be useful, let me know if you'd like it updated!

frej:
> @frej commit message should now be useful, let me know if you'd like it updated!

Much better, thank you for your contribution.

@chrisjbillington: Thanks for the sanity check.

PR #209, opened 2020-03-06
Fix failure to print error message in verify_heads

chrisjbillington:
On Python 3, `b'%s' % None` fails with a TypeError. In verify_heads,
an error message prints the sha1 of a git commit, but that sha1
can be None.

This commit instead prints `b'<None>'` if sha1 is None.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -483,7 +483,7 @@ def verify_heads(ui,repo,cache,force,branchesmap):
     if sha1!=c:
       stderr_buffer.write(
         b'Error: Branch [%s] modified outside hg-fast-export:'
-        b'\n%s (repo) != %s (cache)\n' % (b, sha1, c)
+        b'\n%s (repo) != %s (cache)\n' % (b, b'<None>' if sha1 is None else sha1, c)
       )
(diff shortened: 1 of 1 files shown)


frej:
Good catch, your set of BitBucket repos must be a pretty good testing corpus for fast-export. Thanks for the fix.

chrisjbillington:
Thanks for merging!

Yes, our repos are pretty messy :p. This one arose from an attempted incremental import after the initial import failed on Windows due to two branch names differing only by capitalisation. I thought this would work with `core.ignoreCase false` but apparently not. So will be renaming with a branch mapping file for this repo.

frej:
I think the core.ignoreCase only really does things for filenames, never branch names, i.e. if you do `git add Makefile` and there is a `makefile` it will consider them the same. As branches are represented by files in the file system (named as the branch), git cannot really do anything sensible when two branches are mapped to the same file by the file system. From previous user reports I know that the errors get really confusing when you, for example, have branches named: `Dev/v1.0`, `Dev/v2.0`, and `dev` and try to do the conversion on Windows or OSX :)

PR #210, opened 2020-03-08
Fix bug in ignoring .git files/folders on Windows

chrisjbillington:
Mercurial internally stores (most) filepaths using forward slashes, and
returns them as such from its Python API, even on Windows.

So the splitting up of filepaths with `os.path.sep` was incorrect,
resulting in `.git` files (those within a subdirectory, anyway)
not being ignored on Windows as intended. Splitting on `b'\'` regardless
of OS fixes this.


frej:
Your repos keep on giving :) Thanks for the fix, I tweaked the commit message to say split on `/` instead of `\`to match the code.

chrisjbillington:
Ah, thanks for fixing the typo.

Issue #215, opened 2020-03-23
Convert branch with a new line symbol


fytov:
Hello,

I'm trying to convert a repository (~7k revisions) from hg to git.
I'm having an issue with a branch name in the middle of hg repo(rev. 1500). The name contains a new line symbol (I don't know how it possible, but it's a fact). When commit on this branch is converted (step **Exporting simple delta revision**) I get an error **IOError: [Errno 22] Invalid argument.**

.git/fast_import_crash_<pid> didn't help me. It contains only first part of the branch name (before new line symbol)
Also, I'm using external file with branches names, but I don't know how to specify a new line symbol :(

Traceback:
```
master: Exporting simple delta revision 1835/7203 with 0/1/0 added/changed/removed files
Case 1656, 8785 -
Playlist:Timeline shows twice.: Exporting simple delta revifatal: Branch name doesn't conform to GIT standards: refs/heads/Case 1656, 8785 -
fast-import: dumping crash report to .git/fast_import_crash_15236
sion 1836/7203 with 0/3/0 added/changed/removed files
Traceback (most recent call last):
  File "C:/Users/A/fast-export/hg-fast-export.py", line 692, in <module>
    plugins=plugins_dict))
  File "C:/Users/A/fast-export/hg-fast-export.py", line 544, in hg2git
    plugins)
  File "C:/Users/A/fast-export/hg-fast-export.py", line 349, in export_commit
    export_file_contents(ctx,man,changed,hgtags,fn_encoding,plugins)
  File "C:/Users/A/fast-export/hg-fast-export.py", line 224, in export_file_contents
    wr(d)
  File "C:/Users/A/fast-export/hg-fast-export.py", line 46, in wr
    wr_no_nl(msg)
  File "C:/Users/A/fast-export/fast-export/hg-fast-export.py", line 43, in wr_no_nl
    sys.stdout.write(msg)
```
Just tell me if you need some additional info.

Please feel free to share any ideas.

Alex

chrisjbillington:
You should be able to use escape characters in the branch names file, such as `'\n'`. I believe hg-fast-export's default behaviour is to process the contents of branch mapping entries (same with author mappings) with `.decode('unicode_escape')` before doing anything else with them.

fytov:
> You should be able to use escape characters in the branch names file, such as `'\n'`. I believe hg-fast-export's default behaviour is to process the contents of branch mapping entries (same with author mappings) with `.decode('unicode_escape')` before doing anything else with them.

[...]

PR #217, opened 2020-03-25
Allow utf8 in mappings

chrisjbillington:
We were previously processing entries in mapping files (when
`--mappings-are-raw` is not given) with
`.decode('unicode_escape').encode('utf8')` to replace backslash escape
sequences in bytestrings with the utf-8 encoded characters they
represent. However, it turns out that `.decode
('unicode_escape')` assumes latin-1 encoding if it encounters non-ascii
bytes: https://bugs.python.org/issue21331. So this gave incorrect
results if non-ascii utf8 data was present in the mapping.

To fix this, we now add an extra layer of `.decode('utf8').encode
('unicode-escape')` in order to convert any non-ascii characters into
their backslash escape sequences. Then the subsequent
`.decode('unicode_escape')` only encounters ascii characters and gives
correct results.

This addresses an issue encountered in #215

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,10 +80,10 @@ author information than git, an author mapping file can be given to
 hg-fast-export to fix up malformed author strings. The file is
 specified using the -A option. The file should contain lines of the
 form `"<key>"="<value>"`. Inside the key and value strings, all escape
-sequences understood by the python `string_escape` encoding are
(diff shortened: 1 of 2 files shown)


frej:
Thanks for the fix @chrisjbillington, you're an exemplary contributor!

Issue #222, opened 2020-04-29
Converting a repository with git and hg subrepos fails


ondrej-stanek-ozobot:
`fast-export` can convert git subrepositories records as well as mercurial subrepositories records (when run with appropriate `--subrepo-map` configuration).
However, if a repository contains **both** mercurial and git subrepositories, the conversion fails:
```
Error: <git subrepository name> not found in submodule-mappings
```
The cause is, that in verification phase, `fast-export` falsely expects that both hg and git subrepositories should have appropriate line in the `subrepo-map` file. The case is, that only hg subrepos needs a line in `subrepo-map` that references a converted subrepo, where git subrepositories do not (provided that they are specified as absolute URL).

ondrej-stanek-ozobot:
The issue can be demonstrated by this minimal example:

```
# first, setup a simple repository system for testing
git clone https://github.com/frej/fast-export.git
# checkout the revision on which the bug was reproduced
git -C ./fast-export checkout 61d22307afb4bcbbb76d3e7ff8328e58f8dc714f

hg init ./hg_repo
cd ./hg_repo
# create mercurial subrepo
hg init ./hg_nested 
(cd ./hg_nested && touch ./file && hg add ./file && hg commit -m "commit") # populate it
echo "hg_nested = hg_nested" > .hgsub
hg add .hgsub
hg commit -m "added mercurial subrepo"
echo "fast-export = [git]https://github.com/frej/fast-export.git" >> .hgsub
git clone https://github.com/frej/fast-export.git 
hg commit -m "added fast-export as a git subrepo"
cd -

# next, convert the hg_nested subrepository to git
git init ./git_nested
(cd ./git_nested && ../fast-export/hg-fast-export.sh -r ../hg_repo/hg_nested)

# create the mapping file
echo '"hg_nested"="../git_nested"' > submodule_mappings

# convert the main repository
git init ./git_repo
(cd ./git_repo && ../fast-export/hg-fast-export.sh -r ../hg_repo --subrepo-map=../submodule_mappings)

# review the output of hg-fast-export:
# Error: fast-export not found in submodule-mappings

# cleanup (uncomment)
# rm -rf ./fast-export/ git_nested/ git_repo/ hg_repo/ submodule_mappings
```

ondrej-stanek-ozobot:
We have a fix for the issue that we will submit as PR.

frej:
Fixed in cf0e5837b6c91f4f3f006116cbb36c249b69f58f

PR #223, opened 2020-04-29
Fix issue #222: Converting a repository with git and hg subrepos fails

ondrej-stanek-ozobot:
in verification phase, fast-export falsely expects that both hg and
git subrepositories should have appropriate line in the subrepo-map
file. The case is, that only hg subrepos need a line in subrepo-map
that references a converted subrepo, where git subrepositories do not.
(provided that they are specified as absolute URL)

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -525,14 +525,14 @@ def hg2git(repourl,m,marksfile,mappingfile,headsfile,tipfile,
   	mapping_cache[revnode.encode('hex_codec')] = str(rev)
 
   if submodule_mappings:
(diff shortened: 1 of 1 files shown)


frej:
The change looks good and will eventually be merged, but please review the commit message. Read and apply [this about what constitutes a good commit subject line](https://chris.beams.io/posts/git-commit/#imperative). Also, leave out the issue number, it doesn't tell a person scanning the log very much, besides I'll add it in the commit message for the merge.

frej:
Sorry, I didn't notice the update. Thank you for your contribution!

PR #227, opened 2020-05-12
Add remote of the git repostiory for sub-module

brayra:
The current sub-module support will only add the local mapping to the submodule. This code change will check the submodule for a remote entry. If it exists, it will add that remote, rather than the local mapping to the .gitmodules file

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -139,6 +139,25 @@ def get_author(logmessage,committer,authors):
       return r
   return committer
 
+def git_remote(path):
+    pUrl = r'url\s+=\s+(.*)'
+    configpath = path + b"/.git/config"
+    if os.path.exists(configpath):
+        with open(configpath, 'r') as f:
+            line = f.readline().strip()
+            while line:
+                if line.startswith('[remote'):
+                    line = f.readline().strip()
+                    while not line.startswith('['):
+                        match = re.match(pUrl, line)
+                        if match:
+                            return match.group(1)
+                        line = f.readline().strip()
+                    continue
+                line = f.readline().strip()
+    else:
+        return None
+
 def remove_gitmodules(ctx):
   """Removes all submodules of ctx parents"""
   # Removing all submoduies coming from all parents is safe, as the submodules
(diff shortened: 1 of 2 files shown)


hg-fast-export.py
```diff
@@ -157,7 +176,7 @@ def refresh_git_submodule(name,subrepo_info):
   return b'[submodule "%s"]\n\tpath = %s\n\turl = %s\n' % (name, name, subrepo_info[0])
 
 def refresh_hg_submodule(name,subrepo_info):
-  gitRepoLocation=submodule_mappings[name] + b"/.git"
+  gitRepoLocation=submodule_mappings[name]['path'] + b"/.git"
```
frej, line 179: How will this interact with an existing mappings file for an incremental import?

hg-fast-export.py
```diff
@@ -139,6 +139,25 @@ def get_author(logmessage,committer,authors):
       return r
   return committer
 
+def git_remote(path):
```
frej, line 142: Can't Python's `configparser` understand Git's config format?

brayra: Unfortunately, no. It fails to read it properly

hg-fast-export.py
brayra: This has been used to convert many remote repositories in bitbucket. The mapping file is exactly the same as now. However, the 'path' points to the local copy of the repo and the 'url' points to the remote. 'path' is used to convert without change. However, when the .gitmodules file is written the 'url' is used. 'url' will be same as 'path' if there is no remote.

brayra:
I have added documentation. I hope this makes it clear.

hg-fast-export.py
frej: I took another look, this will probably fly for your example, but changing `load_mapping()` to instead of mapping strings to strings, map strings to dicts instead will pretty much break all other uses of mappings in fast-export (grep for load_mapping, to find all call sites).

hg-fast-export.py
frej: Still on the fence on this, but wouldn't it be better to just fork a git command to ask about this?

frej:
No response to the review comments in more than 90 days, reopen if you intend to address the review comments.

Issue #240, opened 2020-09-02
Fatal: Empty path component found in input


sp-christian:
Crash during import of [GNU Octave](https://hg.savannah.gnu.org/hgweb/octave/).

Most likely with this file path: `gui//Quint.pro`

```
  reset refs/heads/master
  commit refs/heads/master
  mark :13331
  committer Jacob Dawid <jacob.dawid@googlemail.com> 1301955138 +0200
  data 16
  M 100644 inline gui//Quint.pro
* data 415
```

[Crash report](https://gist.github.com/sp-christian/34360eeedaa3a1f68c21086b10ca7ae7)

frej:
Christian <notifications@github.com> writes:

> Most likely with this file path: `gui//Quint.pro`

Looks very likely. I would try writing a plugin which sanitizes the
path. Just change the 'filename' field in the file_data dictionary in
the plugin's file_data_filter callback (See the README for the details).

frej:
No response to review comments for the proposed patch in more than 90 days, closing without prejudice for inactivity.

PR #241, opened 2020-09-09
Add fix_path plugin

bueli:
This plugin sanitizes pathnames containing // (double slash) which will trigger [Empty path component found in input](https://github.com/frej/fast-export/issues/240) issue

### DIFF
--- a/plugins/fix_path/README.md
+++ b/plugins/fix_path/README.md
@@ -1,0 +1,12 @@
+## Fix pathnames
+
+Pathnames in hg repositories can be of the form /folder//subfolder/...
+this results in a crash with the following message:
+```
+fatal: Empty path component found in input
+```
+See [github issue](https://github.com/frej/fast-export/issues/240)
+
+To use the plugin, add
+
+    --plugin fix_path
--- a/plugins/fix_path/__init__.py
+++ b/plugins/fix_path/__init__.py
@@ -1,0 +1,9 @@
+def build_filter(args):
+    return Filter()
+
+class Filter:
+
+    def file_data_filter(self, file_data):
+        filename = file_data['filename']
+        if '//' in filename:
+            file_data['filename'] = filename.replace("//", "/")


plugins/fix_path/__init__.py
```diff
@@ -0,0 +1,12 @@
+def build_filter(args):
+    return Filter(args)
+
+class Filter:
+    def __init__(self, args):
+        i = 1
+
+    # file_data = {'filename':filename,'file_ctx':file_ctx,'d':d}
```
frej, line 8: Why add commented out code?

plugins/fix_path/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+    def file_data_filter(self, file_data):
+        filename = file_data['filename']
+        if filename and '//' in filename:
```
frej, line 11: Why do this conditionally? The filename cannot be empty. Also, why scan the string twice?

plugins/fix_path/README.md
```diff
@@ -0,0 +1,12 @@
+## Fix pathnames
+
+pathnames in hg repositories can be of the form /folder//subfolder/...
```
frej, line 3: Sentences start with a capital letter

plugins/fix_path/__init__.py
frej, line 8: Don't add commented out code

frej (changes requested):
Please take a look at [what is a good commit message](https://chris.beams.io/posts/git-commit/) and amend your message accordingly. Make sure the subject starts with a capital letter and provide at least a one sentence description about what the plugin does in the body.

bueli:
Reduced code and fixed commit message as you requested

plugins/fix_path/__init__.py
```diff
@@ -5,5 +5,5 @@ class Filter:
 
     def file_data_filter(self, file_data):
         filename = file_data['filename']
-        if filename and '//' in filename:
+        if '//' in filename:
```
frej, line 8: You have not answered the question of why you scan the string twice. As far as I can tell the `if '//' in filename` is completely redundant. Am I missing something?

frej (changes requested):
Please squash your commits, there is no reason to preserve the development history of this small change for eternity.

Also make sure that the body of the commit message is wrapped at, at most 80, chars. The [previous linked](https://chris.beams.io/posts/git-commit/) recommends around 72.

Request a re-review when done.

plugins/fix_path/__init__.py
frej: Forget the comment about compiling the regexp, I read the code hastily, sorry.

bueli:
Hi Frej

Will squash soon.

I think your quality goals are very appropriate for production code.
But here we are in a deserted corner of an auxiliary tool ...
Why do you spend and demand so much effort?

You almost lost me, but now i want to understand where the energy is coming from ...

Cheers,
bue

frej:
> But here we are in a deserted corner of an auxiliary tool ... Why do you spend and demand so much effort? You almost lost me, but now i want to understand where the energy is coming from

It's to minimize the maintenance effort, the clearer the patches, the simpler it is to maintain as I can figure out what a change is supposed to be doing. This is a real time saver, as with the declining popularity of Mercurial, I expect the frequency of maintenance actions to become more and more infrequent. The less time I have to spend to remember how something works, the happier I am.

As a maintainer, merging code is always a risk, if I merge a sloppy looking patch, will I have to maintain it in the future, are there hidden bugs? A sloppy looking patch does not really inspire confidence in the quality of the code, especially if the effort to make it "right" in the first place would have been minimal and only would have required a read-through on part of the contributor. Also the willingness to respond to review suggestions and questions is, I have found, a pretty good indication if the contributor will help me if unforeseen problems show up after the merge.

frej:
For the third time, why do you check if `//` occurs in the string before running `filename.replace("//", "/")`, instead of running `replace()` directly? Checking the Python implementation there is no performance benefit from it, you'll end up scanning the string twice when the pattern `//` is in the string.

frej:
No response to review comments in more than 90 days, closing without prejudice for inactivity.

PR #244, opened 2020-09-30
Allow plugins to modify the committer

frej:
Plugins have since they were introduced been able to modify the author
of a commit, but not the committer. This patch adds the necessary
support for allowing them to also modify the committer.


bwh-ct:
Thank, you, this works for me.

frej:
> Thank you, this works for me.

Thanks for the sanity check.

PR #247, opened 2020-10-15
Add plugin for dropping commits from output

lir-ht:
This adds a new plugin for dropping specific commits during conversion, which is useful to get rid of (for example) commits that only added a new tag to .hgtags and will be empty after conversion to git.

The tests require Python 3 but can run the actual program using a different Python version if specified by the PYTHON environment variable, defaulting to the same executable used to run the tests. For example, to run them on Python 2:

	$ PYTHON=./python2-venv/bin/python2 python3 -m unittest

and on Python 3:

	$ ./python3-venv/bin/python3 -m unittest

### DIFF
--- a/plugins/drop/README.md
+++ b/plugins/drop/README.md
@@ -1,0 +1,12 @@
+## Drop commits from output
+
+To use the plugin, add the command line flag `--plugin drop=<spec>`.
+The flag can be given multiple times to drop more than one commit.
+
+The <spec> value can be either
+
+ - a comma-separated list of hg hashes in the full form (40
+   hexadecimal characters) to drop the corresponding changesets, or
+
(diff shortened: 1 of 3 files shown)


frej:
This is a nice example of how to manipulate the commit DAG on the fly and as such I would like to see it merged. 

Could you squash the commits (or at least change the subject of the fixup to have the tag `[plugin/drop]` so it's clear what it refers to), please?

frej:
BTW: @lir-ht, nIce work with the tests, you are the first person ever to have included tests in a PR.

lir-ht:
Thanks! I've squashed the fixup commit as requested.

frej:
Merged, thank you for your contribution @lir-ht!

Grcy3g8 (approved):

PR #249, opened 2020-10-19
add requirements.txt with version for python3.5+

pletnes:
The hg-fast-export script documents support for python 2.7 with
mercurial 4.6; however, python 2.7 is now officially unsupported.

It took me some time to figure out the requirements, which was just one package. I think a `requirements.txt` is the easiest dependency format for a small package like this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+mercurial>=5.2


frej:
The "System Requirements" section in the README already tells the user that fast export depends on Python and Mercurial. I fail to see how adding a subset of this information to `requirements.txt` will make this clearer.

PR #251, opened 2020-11-04
Document --ignore-unnamed-heads option

gillius:
It seems that hg-fast-export can't deal with unnamed heads, even if they are close commits (hg doesn't see these as heads normally). In my repo, I have a handful of abandoned heads that are closed, but hg-fast-export gives errors. On these it is very appropriate to use the --ignore-unnamed-heads option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,11 +198,15 @@ Notes/Limitations
 
 hg-fast-export supports multiple branches but only named branches with
 exactly one head each. Otherwise commits to the tip of these heads
-within the branch will get flattened into merge commits. Chris J
-Billington's [hg-export-tool] can help you to handle branches with
-duplicate heads.
-Alternatively, you can use the [head2branch plugin](./plugins/head2branch)
-to create a new named branch from an unnamed head.
+within the branch will get flattened into merge commits. There are a
+few options to deal with this:
+1. Chris J Billington's [hg-export-tool] can help you to handle branches with
+   duplicate heads.
+2. Use the [head2branch plugin](./plugins/head2branch) to create a new named
+   branch from an unnamed head.
+3. You can ignore unnamed heads with the `-ignore-unnamed-heads` option, which
+   is appropriate in situations such as the extra heads being close commits
+   (abandoned, unmerged changes).
 
 hg-fast-export will ignore any files or directories tracked by mercurial
 called `.git`, and will print a warning if it encounters one. Git cannot
@@ -300,5 +304,12 @@ Frequent Problems
   By design fast export does not touch your working directory, so to
   git it looks like you have deleted all files, when in fact they have
   never been checked out. Just do a checkout of the branch you want.
+
+* `Error: repository has at least one unnamed head: hg r<N>`
+
+  By design, hg-fast-export cannot deal with extra heads on a branch.
+  There are a few options depending on whether the extra heads are
+  in-use/open or normally closed. See [Notes/Limitations](#noteslimitations)
+  section for more details.
 
 [hg-export-tool]: https://github.com/chrisjbillington/hg-export-tool


README.md
```diff
@@ -204,6 +204,13 @@ duplicate heads.
 Alternatively, you can use the [head2branch plugin](./plugins/head2branch)
 to create a new named branch from an unnamed head.
 
+In some projects, heads are closed without being merged (with a close
+branch commit), whether as a way to re-write history or just simply
+abandoned changes. In these cases, it may be appropriate to simply ignore
```
frej, line 209: The first sentence in this paragraph is strange. Why not drop the description of when Mercurial creates multiple heads and just add a sentence to the previous paragraph:

"hg-fast-export can be made to ignore unnamed heads with the `-ignore-unnamed-heads` option.

README.md
```diff
@@ -301,4 +308,11 @@ Frequent Problems
   git it looks like you have deleted all files, when in fact they have
   never been checked out. Just do a checkout of the branch you want.
 
+* Receive `Error: repository has at least one unnamed head: hg r491`
```
frej, line 311: What do you mean with "Receive"? Why not just use the error message "`Error: repository has at least one unnamed head: hg rN`"  as the heading?

frej (changes requested):

README.md
gillius: For some reason I thought I saw this pattern elsewhere in the file. So I've removed that word "receive"

README.md
gillius: My goal was to make it clear the scenario where it is appropriate to use the option. Another case for multiple heads is when you use bookmarks. I don't know if hg-fast-export sees those as branches or not. But, in the case of bookmarks you most definitely do not want to use `-ignore-unnamed-heads`. In the repos I was converting, it was appropriate because all of the extra heads were close commits (which the script does not seem to treat specially).

I agree with combining with earlier paragraph. In my latest commit I combined and re-wrote it to be more like what you proposed.

Ultimately, though, I'm just trying to improve the docs, so I can change it to whatever you want for the PR to complete.

Thanks for working on this script, it is a lot better than the Mercurial plugin and I'm using it to convert multiple repos to git.

README.md
```diff
@@ -308,7 +308,7 @@ Frequent Problems
   git it looks like you have deleted all files, when in fact they have
   never been checked out. Just do a checkout of the branch you want.
 
-* Receive `Error: repository has at least one unnamed head: hg r491`
+* `Error: repository has at least one unnamed head: hg r491`
```
frej, line 308: Perhaps say `Error: repository has at least one unnamed head: hg r<N>` to make it clear that the actual number isn't important.

README.md
frej: The new version is much better. I didn't get that the point you were trying to make was that `-ignore-unnamed-heads` was only appropriate for abandoned and unmerged changes and only then. Breaking it up into a numbered list makes it much clearer!

frej (changes requested):
Please fix my nit-pick about the revision number in the error message, then squash you commits and this is ready for merging.

If you are interested in bookmarks have a look at #61, unfortunately the author abandoned the patch when I pointed out conceptual problems in how the patch translated bookmarks into branches.

README.md
gillius: Done...

gillius:
The git system I use does squashing at the time of completing PRs. I've not done squash with GH PR before. So I squashed my commits and had to "force" push to replace my patch-1 branch. So the old commits are "gone". Hopefully that's what you wanted.

frej:
> The git system I use does squashing at the time of completing PRs. I've not done squash with GH PR before. So I squashed my commits and had to "force" push to replace my patch-1 branch. So the old commits are "gone". Hopefully that's what you wanted.

I don't think Github can do the squash, so what you did is exactly what I wanted.

frej (approved):
Looks good. Thank you for your contribution!

Issue #158, opened 2019-08-20
Instructions in the readme do not work


ilyapopov:
After today's bitbucket announcement of terminating HG repository support on 1 june 2020 I am exploring solutions of converting my hg repositories to git. I am trying to follow instructions given in README, but I encounter the following errors:

```
Error: No --marks option given
```

and then:

```
Error: No --mapping option given
```

These options are not mentioned in readme and I am not sure what to do about these.

ilyapopov:
Sorry, turns out to be my mistake. Accidentally run the `.py` file instead of `.sh`.

frej:
> Therefore, I would suggest to remove that `.py` part in the README
> title.

Patches welcome.

PR #254, opened 2021-01-08
Remove some ".py" wording

rayluo:
Avoid confusion about which file is the main entry point to fast-export, in order to avoid [the issue mentioned here](https://github.com/frej/fast-export/issues/158#issuecomment-754482516), and [agreed here](https://github.com/frej/fast-export/issues/158#issuecomment-754621290).

Also fix a typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-hg-fast-export.(sh|py) - mercurial to git converter using git-fast-import
+hg-fast-export.sh - mercurial to git converter using git-fast-import
 =========================================================================
 
 Legal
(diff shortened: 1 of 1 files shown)


frej (changes requested):
Please write out the issue this patch fixes. Something short like "Avoid confusion about which file is the main entry point to fast-export". That way the git log will be sufficient to understand the reasoning behind this patch, which is good if the project would ever have to move off github.

rayluo:
Thanks for the quick review! I force-pushed the new commit, and also change this PR's description accordingly.

frej:
LGTM, thank you for your contribution @rayluo!

PR #259, opened 2021-02-19
bugfix: Avoid starting incremental conversions from scratch

### LINKED ISSUE #258: fast-export does not follow last stored state
We are doing incremental synchronization between our mercurial and git repositories. I was moving this job to different server today and it stopped working. It was always going from start. At the end I found this problem:

```python
  min=int(state_cache.get('tip',0)) # Here it is always 0
  max=_max
  if _max<0 or max>tip:
    max=tip
```

It turned out that state_cache is actually having byte strings which does not match 'tip'
```
{b'repo': b'../hgrepository/', b'tip': b'18544'}
```

After I changed it to this it works again:
```python
  min=int(state_cache.get(b'tip',0)) # Notice that b
```

I am not sure why the behavior changed (maybe python version), but I think it should be searched as byte string as load_cache it creating byte strings.

frej:
Keys and values in the state cache are byte strings, therefore a
lookup of 'tip' will always fail. The failure makes the conversion
start over from the beginning, but as fast-export is deterministic the
results are the same, just very inefficient. The bug has existed since
the port to Python 3.

This patch switches the 'tip' lookup to use a byte string which should
make incremental conversions restart at the last converted commit. As
'x' == b'x' in Python 2, this should be a backwards compatible change.

Bug reported and fix suggested by Tomas Kolda.

Fixes #258.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -547,7 +547,7 @@ def hg2git(repourl,m,marksfile,mappingfile,headsfile,tipfile,
   except AttributeError:
     tip=len(repo)
 
-  min=int(state_cache.get('tip',0))
+  min=int(state_cache.get(b'tip',0))
   max=_max
   if _max<0 or max>tip:
     max=tip
@@ -580,8 +580,8 @@ def hg2git(repourl,m,marksfile,mappingfile,headsfile,tipfile,
     for rev in range(min,max):
       c=export_note(ui,repo,rev,c,authors, encoding, rev == min and min != 0)
 
-  state_cache['tip']=max
-  state_cache['repo']=repourl
+  state_cache[b'tip']=max
+  state_cache[b'repo']=repourl
   save_cache(tipfile,state_cache)
   save_cache(mappingfile,mapping_cache)
 


frej:
@chrisjbillington : This seem Python 2 vs 3 related, could you take a look?

chrisjbillington:
Apologies for introducing the bug!

This fix should be fine, though it seems like possibly the more brittle option since all lookups of the cache dict will need to specify bytestrings in the future.

The other alternative is to have the cache dict's keys be unicode strings:

```diff
diff --git a/hg2git.py b/hg2git.py
index e94270b..14dfb07 100755
--- a/hg2git.py
+++ b/hg2git.py
@@ -113,8 +113,8 @@ def load_cache(filename,get_key=mangle_key):
     if fields==None or not len(fields)==2 or fields[0][0:1]!=b':':
       sys.stderr.write('Invalid file format in [%s], line %d\n' % (filename,l))
       continue
-    # put key:value in cache, key without ^:
-    cache[get_key(fields[0][1:])]=fields[1].split(b'\n')[0]
+    # put key:value in cache, key without ^. Keys are unicode strings, values are bytes:
+    cache[get_key(fields[0][1:]).decode('utf8')]=fields[1].split(b'\n')[0]
   f.close()
   return cache
```

Since the keys are always strings chosen by hg-fast-export and not user data, there's no need for them to be bytestrings. And since `u'x' == 'x'` in Python 2, lookups with bytestrings like `cache['tip']` will succeed in Python 2 even though the keys would be unicode.

Either solution works, so it's a matter of taste I think.

frej:
> Apologies for introducing the bug!

No need to apologize, contributing the Python 3 port has substantially increased your quota of allowed bugs :) Thanks for the review.

> Either solution works, so it's a matter of taste I think.

As `load_cache()` is also used to load the branch mappings file, where keys are user supplied, I'm a bit hesitant to change it to enforce utf8. Right now a user can remap a branch name that isn't utf8, or even broken utf8 (Mercurial apparently accepts anything). So I think having to remember to specify byte strings is an ok tradeoff.

chrisjbillington:
Ah yes, if the same structures are used for arbitrary user data then it should be bytestrings, and better to have it always be bytestrings than sometimes one and sometimes the other. So I agree!

Issue #257, opened 2021-02-14
Feature Request: Read and store hg-subversion markers during conversions


SirIntellegence:
My company has a repository that was a simple svn repo for which I made a mercurial mirror (using hg-subversion) at some point in time. I am now thinking of using git so I can use https://github.com/bozaro/git-as-svn for reasons. The mercurial mirror has markers in it for what the original svn revision number is when it was converted. I would really like it if that could be imported into the metadata around where one of the plugins copies the hg hash if possible.

frej:
I'm not really sure what you are asking for. Are you asking about a new feature adapted to the output of your (unspecified) SVN to Mercurial conversion? Or are you asking for the addition of additional information (again unspecified) to be exposed in the plugin API? If it is the latter, please tell me what information that is lacking, or even better, send a patch.

Custom mangling of commit messages is something the plugin system was designed to do. Unless the tool you have used for the initial conversion from SVN and git-as-svn are widely used, I'd say that such a feature is of too niche to be maintained together with fast-export.

SirIntellegence:
I mostly mean the svn revision markers created by https://www.mercurial-scm.org/wiki/HgSubversion. The extension itself seems to be sorely outdated and unmaintained (see https://stackoverflow.com/a/64563930/1366594), so I understand if you don't want to implement that. What would be nice is to be able to grab a list of unclaimed properties (or something like that) for each revision so I don't have to figure out how to invoke mercurial to get the property for that revision.

frej:
After some quick Googling it looks like Mercurial does not natively support the kind of properties you are talking about. I very much suspect that the properties you are talking about comes from the HgSubversion extension you mentioned, but I could be wrong. Feel free to submit a patch if there are such generic properties (not depending on a particular extension).

If you don't want to spawn a hg command for each revision and the properties depend on an extension, you could traverse the complete history in Mercurial, dump a list of `<hash>:<property>` and then feed that to your plugin.

SirIntellegence:
Ok. I did some research and found this in case it helps: https://stackoverflow.com/q/4268745/1366594

frej:
Feel free to submit a patch that exposes the `extra`-field to plugins.

frej:
Fix merged in #260

PR #260, opened 2021-02-19
Suggested changes to add plugin support for the 'extra' field

SirIntellegence:
Seems to work in my copy, but I am not a python programmer, so you may want to double check it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ defined filter methods in the [dos2unix](./plugins/dos2unix) and
 [branch_name_in_commit](./plugins/branch_name_in_commit) plugins.
 
 ```
-commit_data = {'branch': branch, 'parents': parents, 'author': author, 'desc': desc, 'revision': revision, 'hg_hash': hg_hash, 'committer': 'committer'}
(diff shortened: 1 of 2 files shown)


SirIntellegence:
See #257

frej (changes requested):
Oh, it was that simple!

Please amend your commit to update the **Plugins** section in `README.md` to document the change, also take a look at the **Submitting Patches** section and make sure that your commit message follows Chris Beams' suggestions.

frej (changes requested):
Why not just change the commit subject to "Add plugin support for the 'extra' field"? We are way past suggestions now, this is a patch I want to merge! I could make the edit, but as your name will be associated with the final result, I prefer not to do it, as it could be considered stepping on your toes.

SirIntellegence:
Ah, didn't think about that. Just a moment...

SirIntellegence:
Also, for reference, I currently can only confirm that it doesn't crash. I haven't tried reading "extra" yet.

SirIntellegence:
I just recently tested it and it does work properly

frej:

frej (approved):

PR #262, opened 2021-03-05
Added ability for file filter plugins to discard files

Dakkaron:
This pull request allows file data filter plugins to discard files.

For example, if you use this plugin:

```
def build_filter(args):
    return Filter(args)

class Filter:
    def __init__(self, args):
        self.whitelistedPrefixes = args.split(",")

    def file_data_filter(self,file_data):
        filename = file_data['filename']
        file_data["drop_file"] = not any([ filename.startswith(x) for x in self.whitelistedPrefixes ])

```

and call it with the following parameters: `--plugin file_prefix_whitelist='src/'`

only files in the sub-folder `src/` will be included.

This might create some empty commits, but that doesn't cause any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,8 @@ can be modified by any filter. `file_ctx` is the filecontext from the
 mercurial python library.  After all filters have been run, the values
 are used to add the file to the git commit.
 
+To discard a file, set the `"drop_file"` value in its `file_data` to `True`.
+
 Submodules
 ----------
 See README-SUBMODULES.md for how to convert subrepositories into git
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -231,7 +231,9 @@ def export_file_contents(ctx,manifest,files,hgtags,encoding='',plugins={}):
       d=file_data['data']
       filename=file_data['filename']
       file_ctx=file_data['file_ctx']
-
+      if file_data.get('drop_file',False):
+        continue # Plugin dropped the file change
+    
     wr(b'M %s inline %s' % (gitmode(manifest.flags(file)),
                            strip_leading_slash(filename)))
     wr(b'data %d' % len(d)) # had some trouble with size()
--- a/plugins/sync_files_by_prefix/README.md
+++ b/plugins/sync_files_by_prefix/README.md
@@ -1,0 +1,10 @@
+## Whitelist files and folders
+
+To use the plugin, add the command line flag `--plugin file_prefix_whitelist=<prefixes>`.
+<prefixes> is a comma separated list of prefixes. Files with a path that doesn't start
+with any of the prefixes will be discarded. This can create empty commits.
+
+Examples for prefixes:
+ - "src/": This prefix whitelists everything under the src folder.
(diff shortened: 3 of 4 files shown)


frej (changes requested):
Please read and follow the "Submitting Patches"-section in the readme.

Additionally for this patch, I would expect an update to the "Plugins"-section as it extends the plugin API. As you have created an example for the PR description, add it to the set of examples in `plugins/`

Dakkaron:
I hope this is better now

frej (changes requested):
Much better.

Please [use the imperative mood in the subject line](https://chris.beams.io/posts/git-commit/#imperative) as Chris Beams' recommends.

Although I haven't converted the name of the main development branch in this repo, I think we should avoid the term _whitelist_ following the Linux kernel's example. Why not name the example `filter_by_path_prefix` and rename `whitelistedPrefixes` to `wantedPrefixes` to side step the whole issue?

Dakkaron:
`filter_by_path_prefix` would be a useful term, but this project uses the word `filter` for a concept that really doesn't have much to do with filtering, but instead uses it to mean something like modifier. So naming this a filter would be rather misleading.

The issue with the word `whitelist` is that it really doesn't have anything to do with racism, but is instead a really descriptive term for exactly what is meant here.

Tbh, this pull request has 3 lines of useful code. I could have just made an issue with a feature request, now I'm slogging through a racism debate for some example code. If you want to, lets just delete the example and people can figure out how to use that thing themselves.

frej:
> `filter_by_path_prefix` would be a useful term, but this project uses the word `filter` for a concept that really doesn't have much to do with filtering, but instead uses it to mean something like modifier. So naming this a filter would be rather misleading.

Good point, that's true, the methods in the plugin API should really be renamed, but I don't think anyone will be confused. At least I haven't received any issues on that aspect of plugin use. Git calls the corresponding functionality in `filter-branch` (which could be an alternative for one-off conversions) a _subdirectory-filter_ (it also has _index-filter_).

> The issue with the word `whitelist` is that it really doesn't have anything to do with racism, but is instead a really descriptive term for exactly what is meant here.

This is the kind of debate I wanted to avoid by my suggestion to not use a word that I know is a trigger for certain activists.

> Tbh, this pull request has two lines of useful code. I could have just made an issue with a feature request, now I'm slogging through a racism debate for some example code. If you want to, lets just delete the example and people can figure out how to use that thing themselves.

It's up to you to decide on what to do, either address the review comment in question or drop the plugin example.

frej (changes requested):
`sync_file_by_prefix`? If you claim that my suggestion of `filter_by_path_prefix` could confuse users, how do you get "sync" to mean "only keep files matching"? 

The updated patch seems half-baked, for example `sync_files_by_prefix/README.md` is not updated. I understand that you are reluctant to change your patch, but in case you can't accept my suggestions for naming, please just drop the example. As you said, this is a ridiculous amount of work for a patch adding two-lines of "real" code.

frej:
No activity in more than 90-days, closing.

PR #263, opened 2021-03-16
Tab char added to sanitizing regex

poterba:


frej:
Rejected. The comment at the start of the function this patch modifies explains why.

Issue #32, opened 2014-07-04
Fail on old repository


stoune:
I have quite large repository. When I'm exporting to git from hg I'm receiving error:

```
master: Exporting thorough delta revision 2804/4207 with 0/23/1 added/changed/removed files
Traceback (most recent call last):
  File "../fast-export/hg-fast-export.py", line 432, in <module>
    options.statusfile,authors=a,sob=options.sob,force=options.force,hgtags=options.hgtags,notes=options.notes))
  File "../fast-export/hg-fast-export.py", line 362, in hg2git
    c=export_commit(ui,repo,rev,old_marks,max,c,authors,sob,brmap,hgtags,notes)
  File "../fast-export/hg-fast-export.py", line 224, in export_commit
    export_file_contents(ctx,man,changed,hgtags)
  File "../fast-export/hg-fast-export.py", line 136, in export_file_contents
    wr(d)
  File "../fast-export/hg-fast-export.py", line 37, in wr
    wr_no_nl(msg)
  File "../fast-export/hg-fast-export.py", line 34, in wr_no_nl
    sys.stdout.write(msg)
IOError: [Errno 32] Broken pipe
```

I'm used following options:
 ../fast-export/hg-fast-export.sh -r reponame/ --hgtags --hg-hash --force

lwcolton:
@frej This is not a duplicate of #34 .  If you look at the call stack here, sanitize_name is not being called at all.  I am experiencing this issue as well and will provide a pull request if I can.

danielpunkass:
I just ran into this with one of my projects. Same stack trace with export_file_contents, etc. No sanitize_name in the stack.

danielpunkass:
In my case the crash was occuring in fast import because one of my branch names was problematic, even after being sanitized by sanitize_name. The imported repository had a branch named erroneously after a whole ssh:// URL, so something like:

ssh://daniel@where-ever.com/repos/something

The branch name was sanitized to replace the colon, but apparently the "/" also causes problems. I'll submit a pull request with a tweak to add / to the regex for sanitizing. This worked around the issue for me.

danielpunkass:
Anybody else who comes across this issue, the right solution is to use the branch renaming -B flag to alleviate any problems that may be occurring because of the old Mercurial branch names. See explanation on #285.

PR #285, opened 2022-02-26
Add forward slash to the list of characters to sanitize in sanitize_name

danielpunkass:
See Issue #32, I believe this is the correct fix for the underlying issue causing that import failure, or at least it addresses one of the possible branch naming problems that could lead to the issue.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -266,7 +266,7 @@ def sanitize_name(name,what="branch", mapping={}):
   if not auto_sanitize:
     return mapping.get(name,name)
   n=mapping.get(name,name)
-  p=re.compile(b'([\\[ ~^:?\\\\*]|\.\.)')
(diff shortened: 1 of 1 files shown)


frej:
Please read the comment block starting the function your are modifying. It both contains an explanation for why your patch is the wrong solution and what the right solution to the problem which triggered your PR is. Besides that, `/` is perfectly valid in Git branch names, so there is no reason to sanitize it. Read the [first item](https://github.com/frej/fast-export#frequent-problems)  in the  "Frequent Problems" section of the README for a crash course on Git branch naming.

danielpunkass:
Oh! I'm embarrassed to have glossed right over that comment in the sources. Thanks for explaining so patiently.

paulwratt:
:)

"live and learn"

PR #291, opened 2022-09-20
Update README.md to change recommandation for crlf filtering

nicolas-van:
Changed the recommended `crlf-filter.sh` to add the `-q` option to `dos2unix` call to avoid failing on some rare file encodings.

Tested on a 10-years old humongously big mercurial repository with a mix of everything you can imagine in file encodings. Worked as expected.

Also changed the comments a little bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,10 @@ is to convert line endings in text files from CRLF to git's preferred LF:
 # $2 = Mercurial's hash of the file
 # $3 = "1" if Mercurial reports the file as binary, otherwise "0"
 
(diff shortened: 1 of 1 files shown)


frej (changes requested):
Running `dos2unix` with `-q` seems like the right thing to do, I approve that change. But please don't change the way example files are displayed in the readme as that is unrelated to the patch subject (and would make the formatting inconsistent) .

nicolas-van:
> Please squash the commits.

Done

frej:

PR #295, opened 2023-03-02
Fix data handling

felipec:
The length should be exactly the same.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -197,7 +197,7 @@ def refresh_gitmodules(ctx):
 
   if len(gitmodules):
     wr(b'M 100644 inline .gitmodules')
-    wr(b'data %d' % (len(gitmodules)+1))
+    wr(b'data %d' % (len(gitmodules)))
     wr(gitmodules)
 
 def export_file_contents(ctx,manifest,files,hgtags,encoding='',plugins={}):
@@ -315,7 +315,7 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
     parents = commit_data['parents']
     author = commit_data['author']
     user = commit_data['committer']
-    desc = commit_data['desc']
+    desc = commit_data['desc'] + b'\n'
 
   if len(parents)==0 and revision != 0:
     wr(b'reset refs/heads/%s' % branch)
@@ -325,9 +325,8 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
   if sob:
     wr(b'author %s %d %s' % (author,time,timezone))
   wr(b'committer %s %d %s' % (user,time,timezone))
-  wr(b'data %d' % (len(desc)+1)) # wtf?
+  wr(b'data %d' % (len(desc)))
   wr(desc)
-  wr()
 
   ctx=revsymbol(repo, b"%d" % revision)
   man=ctx.manifest()
@@ -389,7 +388,7 @@ def export_note(ui,repo,revision,count,authors,encoding,is_first):
   wr(b'N inline :%d' % (revision+1))
   hg_hash=revsymbol(repo,b"%d" % revision).hex()
   wr(b'data %d' % (len(hg_hash)))
-  wr_no_nl(hg_hash)
+  wr(hg_hash)
   wr()
   return checkpoint(count)
 
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -55,6 +55,10 @@ def wr(msg=b''):
   stdout_buffer.write(b'\n')
   #map(lambda x: sys.stderr.write('\t[%s]\n' % x),msg.split('\n'))
 
+def wr_data(data):
+  wr(b'data %d' % (len(data)))
+  wr(data)
+
 def checkpoint(count):
   count=count+1
   if cfg_checkpoint_count>0 and count%cfg_checkpoint_count==0:
@@ -197,8 +201,7 @@ def refresh_gitmodules(ctx):
 
   if len(gitmodules):
     wr(b'M 100644 inline .gitmodules')
-    wr(b'data %d' % (len(gitmodules)))
-    wr(gitmodules)
+    wr_data(gitmodules)
 
 def export_file_contents(ctx,manifest,files,hgtags,encoding='',plugins={}):
   count=0
@@ -325,8 +328,7 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
   if sob:
     wr(b'author %s %d %s' % (author,time,timezone))
   wr(b'committer %s %d %s' % (user,time,timezone))
-  wr(b'data %d' % (len(desc)))
-  wr(desc)
+  wr_data(desc)
 
   ctx=revsymbol(repo, b"%d" % revision)
   man=ctx.manifest()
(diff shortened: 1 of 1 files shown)


frej (commented):
Please have a look at what constitutes a [good commit message](https://cbea.ms/git-commit/), I'm particularity missing why this change is necessary "Fix data handling" doesn't tell me much.

felipec:
> Please have a look at what constitutes a [good commit message](https://cbea.ms/git-commit/)

I have hundreds of commits in [git.git](https://git.kernel.org/pub/scm/git/git.git/log/?qt=author&q=felipe.contreras), I think I know what a good commit message is.

> I'm particularity missing why this change is necessary "Fix data handling" doesn't tell me much.

That's why I added the explanation "The length should be exactly the same".

What does that mean?

It means the length of `data %d` should be the actual  length of the data, not `len(gitmodules)+1`, that's a **hack**.

The removed comment that is present since day 1 suggest it was always a hack:

    wr(b'data %d' % (len(desc)+1)) # wtf?
    
It should be `len(desc)` (exactly the same). I thought that part was clear.

The only reason the hack works, is that after the data is written, an extra `wr()` is issued, essentially adding `\n` to the stream. But this is wrong, the extra new line should be added to the data (desc).

The reason why that wtf comment is there is that commit messages in git typically end with a new line, therefore the `wr()` replicated what a typical commit message looks like in the stream, but it doesn't align with the original data (mercurial commits don't have that).

Adding an extra b'\n' to `desc` does the same thing, but correctly.

BTW, I'm the author of `git-remote-hg`, I fixed that problem in 2013, and the output is compatible with `hg-git`, who fixed the problem in 2009: [added a missing newline back to git conversion](https://foss.heptapod.net/mercurial/hg-git/-/commit/0e0a2d20deed9552df41153fde587a8145b1c64d).

I'm not so certain about the `.gitmodules`, but I bet git adds a newline, which is why the `+1` is there, but I don't see you adding a newline in the stream, so presumably this screws up the stream, and thus there's a bug.

Does this explanation answer you questions?

hg-fast-export.py
```diff
@@ -196,8 +196,9 @@ def refresh_gitmodules(ctx):
       gitmodules+=refresh_hg_submodule(name,subrepo_info)
 
   if len(gitmodules):
+    gitmodules += b'\n'
```
frej, line 199: This is redundant (see the review comment)

frej (commented):
> Does this explanation answer you questions?

Yes that explains the why and what of this commit. 

From that description it is clear why this change is the right thing to do. As you clearly know how to write good commit messages (I had a gander at your 20 latest commits in git.git and they clearly described what and why something was being fixed), it's unfortunate that you did not apply the same care to this PR.

May I suggest that you amend the commit (or if you OK it, I can do it) to something along the lines of:

_Hg-fast-export has since the first version fudged the encoding of the length of commit messages by adding 1 to the message length and then outputting an extra line feed following the message contents (using `wr()`). This commit changes the logic to add a trailing `\n` to the commit message and can thus remove the extra `wr()` and the `+1` length adjustment._

_Also adjust the output of `.gitmodules` to not drop the optional `\n` after a `data` command._

> I'm not so certain about the `.gitmodules`, but I bet git adds a newline, which is why the `+1` is there, but I don't see you adding a newline in the stream, so presumably this screws up the stream, and thus there's a bug.

`wr()` always adds a newline, so the `+1` is harmless as it "moves" the optional `\n` after a `data` command to the contents of  `.gitmodules`. As the contents of `.gitmodules` is generated to end with a `\n`, you can drop the `gitmodules += b'\n'`.

hg-fast-export.py
felipec: It's not redundant because it allows the removal of the `+1` in `len(gitmodules)+1`, but then `wr()` adds an extra newline, when it should be `wr_no_nl()`.

I've fixed this and pushed a new pull request.

felipec:
> As the contents of `.gitmodules` is generated to end with a `\n`, you can drop the `gitmodules += b'\n'`.

I actually looked at the data and the newline is already included so there wasn't any need to add the `\n`, or use `wr()`: it should be `wr_no_nl()`.

hg-fast-export.py
```diff
@@ -55,6 +55,10 @@ def wr(msg=b''):
   stdout_buffer.write(b'\n')
   #map(lambda x: sys.stderr.write('\t[%s]\n' % x),msg.split('\n'))
 
+def wr_data(data):
+  wr(b'data %d' % (len(data)))
+  wr_no_nl(data)
```
frej, line 59: I like `wr_data()`, but i don't like dropping the optional `\n`, as the git fast-import` manpage says:

_The LF after <raw> is optional (it used to be required) but recommended. Always including it makes debugging a fast-import stream easier as the next command always starts in column 0 of the next line, even if <raw> did not end with an LF._

hg-fast-export.py
```diff
@@ -51,8 +51,7 @@ def wr_no_nl(msg=b''):
     stdout_buffer.write(msg)
 
 def wr(msg=b''):
-  wr_no_nl(msg)
-  stdout_buffer.write(b'\n')
+  wr_no_nl(msg + b'\n')
```
frej, line 54: My Python intuition is not developed enough to know if two writes is more or less efficient than doing an append when the prefix is large. Do you know?

hg-fast-export.py
frej: The extra `\n` is not bad, it increases the readability of the dump. And in this case it saved us when when we told git that the data was one byte larger than it was.

hg-fast-export.py
felipec: Should make no difference. In Python 3 `stdout_buffer` is a BufferedWriter anyway, so `write()` is not actually writing, just adding to the buffer.

hg-fast-export.py
felipec: But it's not just adding a newline on the stream, it's adding a newline on the file. If you want an extra newline, then probably `wr()` should be used instead of `wr_no_nl()`, but you can't add arbitrary newlines, care must be taken to only add one. Which is why `wr_data()` makes sense.

hg-fast-export.py
felipec: Fair enough. The good thing about the helper is that it's updated everywhere at once.

frej:
> > As the contents of `.gitmodules` is generated to end with a `\n`, you can drop the `gitmodules += b'\n'`.
> 
> I actually looked at the data and the newline is already included so there wasn't any need to add an extra `\n`, or use `wr()`: it should be `wr_no_nl()`.

For readability it is, especially when we go to `wr_data()`.

felipec:
The updated `wr_data()` would be readable.

hg-fast-export.py
frej: > it's adding a newline on the file

Yes, that's why I said it was redundant as the file already ends with a `\n`. We drop the `+1` here, and the new `wr_data()` correctly writes the data and preserves the readability. The extra `\n` added by `wr()` was what saved us from the effects of the erroneous `+1` as the inadvertently  added `\n` to `.gitmodules` is harmless and fast-import accepted our output despite it being conceptually wrong.

frej (approved):
The commit message for "Trivial simplification of wr" should be:

"No need to issue _two_ write commands."

I'll tweak it and merge this PR. Thank you for your contribution!

frej:
Merged in 4edea927fb94e2b1ee9d52cb61a12a59fc60f5d2

PR #296, opened 2023-03-04
README: cleanup initial instructions

felipec:
The `git init` command can create the directory, and HEAD doesn't need to be specified in `git checkout` (it's the default).


frej:
Merged, thank you for the PR.

PR #298, opened 2023-03-14
Add tests

felipec:
It would be useful to have some kind of tests, I personally like [Sharness](http://felipec.github.io/sharness/) (which I recently inherited).

It uses shell scripts and tests are really simple:

```sh
test_expect_success 'foo' '
  hg init
'
```

In this proposal I've added just two tests, but the merge one is already useful to detect issues in some changes I'm working on.

This is what I use in `git-remote-hg` which has quite extensive tests (e.g. [main.t](https://github.com/felipec/git-remote-hg/blob/master/test/main.t)), also what I'm using in my new project `hg-fast-export` (e.g. [main.t](https://github.com/felipec/hg-fast-export/blob/master/t/main.t)), and this is what the Git project uses (e.g. [t0001-init.sh](https://github.com/felipec/hg-fast-export/blob/master/t/main.t)).

It's pretty easy to setup, one can just clone the repository to `t/sharness`, or we can include the whole script which is around 760 lines of code.

It's also easy to setup GitHub actions to run these tests, as you can see I'm setting up an action in this pull request, and the results of one job are [here](https://github.com/felipec/hg-fast-export-frej/actions/runs/4411322327/jobs/7729727359).

It would be nice to have at least some tests, and even nicer if we could share them between projects.

### DIFF
--- a/t/main.t
+++ b/t/main.t
@@ -1,0 +1,94 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 Felipe Contreras
+#
+
+test_description='Main tests'
+
+. "${SHARNESS_TEST_SRCDIR-/usr/share/sharness}"/sharness.sh || exit 1
+
+check() {
+	echo "$3" > expected &&
+	git -C "$1" show -q --format='%s' "$2" > actual &&
+	test_cmp expected actual
+}
+
+git_clone() {
+	(
+	git init -q "$2" &&
+	cd "$2" &&
+	hg-fast-export.sh --repo "../$1"
+	)
+}
+
+setup() {
+	cat > "$HOME"/.hgrc <<-EOF
+	[ui]
+	username = H G Wells <wells@example.com>
+	EOF
+}
+
+setup
+
+test_expect_success 'basic' '
+	test_when_finished "rm -rf hgrepo gitrepo" &&
+
+	(
+	hg init hgrepo &&
+	cd hgrepo &&
+	echo zero > content &&
+	hg add content &&
+	hg commit -m zero
+	) &&
+
+	git_clone hgrepo gitrepo &&
+	check gitrepo @ zero
+'
+
+test_expect_success 'merge' '
+	test_when_finished "rm -rf hgrepo gitrepo" &&
+
+	(
+	hg init hgrepo &&
+	cd hgrepo &&
+	echo a > content &&
+	echo a > file1 &&
+	hg add content file1 &&
+	hg commit -m "origin" &&
+
+	echo b > content &&
+	echo b > file2 &&
+	hg add file2 &&
+	hg rm file1 &&
+	hg commit -m "right" &&
+
+	hg update -r0 &&
+	echo c > content &&
+	hg commit -m "left" &&
+
+	HGMERGE=true hg merge -r1 &&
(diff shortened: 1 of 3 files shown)


t/main.t
```diff
@@ -0,0 +1,94 @@
...
+	HGMERGE=true hg merge -r1 &&
+	hg commit -m "merge"
+	) &&
+
+	git_clone hgrepo gitrepo &&
+
+	cat > expected <<-EOF &&
+	left
+	c
+	tree @:
```
frej, line 78: What does `@` do here? I can't find it in the API documentation for Sharness, and it's not a part of shell-script I am familiar with.

frej:
This looks very appealing, I like the idea of more testing. I have been using a (not published) smoke test to check for breaking changes. I'll take a stab at converting it to Sharness. If that works out, I'll merge this one (and then #299 to verify that it doesn't change the result).

> It's pretty easy to setup, one can just clone the repository to `t/sharness`, or we can include the whole script which is around 760 lines of code.

I'd prefer to add it as a submodule, that way there is never a question about which version of the framework which is supposed to be used and it is fairly easy to see if there are upstream changes.

> It would be nice to have at least some tests, and even nicer if we could share them between projects.

I agree. If the common tests were in their own repo we could just include them as a submodule.

t/main.t
felipec: It's not sharness, it's git: `git show @:` is the equivalent of `git show HEAD:`, which means "show the tree of the current branch".

felipec:
> I'd prefer to add it as a submodule, that way there is never a question about which version of the framework which is supposed to be used and it is fairly easy to see if there are upstream changes.

I'm not a fan of submodules, but that's possible.

> I agree. If the common tests were in their own repo we could just include them as a submodule.

That's also a possibility. I've been running some tests interchangeably with your `hg-fast-export`, my `hg-fast-export`, and Mercurials official `hg fastexport`.

t/main.t
frej: Ah, then I learnt something new today.

felipec:
Any specific reason not to use `make` to run the tests?

I checked your `run-tests` script which basically just executes the test files in `t/` and if `parallel` is found, use it.

But the same can be dune by using `make` (if you have `MAKEFLAGS=-j`). I copied the example `Makefile` in sharness' [example/Makefile](https://github.com/felipec/sharness/blob/master/example/Makefile), and it works fine.

BTW. It's a good idea to remove the sharness cruft before and after the tests. See the Makefile.

felipec:
Also, what is this about?

```sh
if ! $READLINK -f "$(which "$0")" > /dev/null 2>&1 ; then
    ROOT="$(dirname "$(which "$0")")"
else
    ROOT="$(dirname "$($READLINK -f "$(which "$0")")")"
fi
```

I suppose you want the absolute path of the real location of the project, but you can do that in a portable way with: `$(cd "$(dirname "$0")" && pwd -P)`.

frej:
`$(cd "$(dirname "$0")" && pwd -P)` is nicer than the old stuff we've been using in `hg-fast-export.sh` to locate the python files.

And using make is a nicer way too.

PR #299, opened 2023-03-15
Simplifications of get_filechanges

felipec:
This code hasn't been touched in decades, but most of it is not necessary. In [git-remote-hg](https://github.com/felipec/git-remote-hg/blob/master/git-remote-hg#L316) we've been using a simplified version of `get_filechanges`+`split_dict` for many years (maybe a decade?).

All that is needed is to compare the files of the new commit with the ones of the first parent.

To test the different versions of the code I wrote a benchmarking tool [filechanges-perf](https://github.com/felipec/hg-fast-export/blob/tool/filechanges-perf/tools/filechanges-perf). The new version of the code is reliably faster, and produces the same resulting repos. In the case of `hg` it's almost twice as fast.

I tested with `hg`, `mozilla-central`, and `hg-git`:

```
hg '0:10000':
3cd697c5eccb477a29f9570d2b829d8ee807b0cc
0: 46.59
3cd697c5eccb477a29f9570d2b829d8ee807b0cc
1: 25.71

mozilla-central '0:10000':
c027d5531d03c29b879ca69f58d3e8997e8ae7e8
0: 113.60
c027d5531d03c29b879ca69f58d3e8997e8ae7e8
1: 102.37

hg-git '0:tip':
a4c5b8d1e35739501a7141b844fc4eaea57b6c5a
0: 3.25
a4c5b8d1e35739501a7141b844fc4eaea57b6c5a
1: 2.63
```

Same output, faster execution, and simpler code.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -74,35 +74,34 @@ def file_mismatch(f1,f2):
   """See if two revisions of a file are not equal."""
   return node.hex(f1)!=node.hex(f2)
 
-def split_dict(dleft,dright,l=[],c=[],r=[],match=file_mismatch):
+def split_dict(dleft,dright,c=[],r=[],match=file_mismatch):
   """Loop over our repository and find all changed and missing files."""
   for left in dleft.keys():
     right=dright.get(left,None)
     if right==None:
-      # we have the file but our parent hasn't: add to left set
-      l.append(left)
+      # we have the file but our parent hasn't: add
+      c.append(left)
     elif match(dleft[left],right) or gitmode(dleft.flags(left))!=gitmode(dright.flags(left)):
-      # we have it but checksums mismatch: add to center set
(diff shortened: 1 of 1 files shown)


frej:
I like this and also #298. I will merge #298 first, then convert my (not published) smoke test to Sharness, merge that and then merge this PR. If this PR results in NFC, the smoke-test should produce the same result.

frej:
@felipec: I have now merged both @298 and #299, thank you for your contribution!

felipec:
Great! I presume you tried to clone different repositories and the result was exactly the same.

frej:
Yes, I tested with the gmp,  hg,  lemon,  nginx,  pidgin, and  xine-lib repos.  `git fast-export --all | sha1sum` produces the same hash for both versions.

PR #311, opened 2023-11-11
Make pluginloader use importlib instead imp

### LINKED ISSUE #310: Support for Python 3.12 (e.g. replace `imp` usage for Python 3 users)
When migrating some `mercurial`-dependent formulae in Homebrew to Python 3.12 (https://github.com/Homebrew/homebrew-core/pull/152277), the CI found that this package is incompatible with Python 3.12 at least due to usage of removed `imp`
https://github.com/frej/fast-export/blob/0d0e90d3280e83580e5531d70ca54450b9593a89/pluginloader/__init__.py#L2

Removal documented at https://docs.python.org/3/whatsnew/3.12.html#imp along with recommended `importlib` equivalent.

`imp` has been deprecated since Python 3.4. Given documented support for Python 3.5+, it is probably ideal to use supported `importlib` module when running Python 3.

onlined:
Python 3.12 has removed `imp` and it's recommended to use `importlib` instead.

Closes #310.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,9 @@ first time.
 System Requirements
 -------------------
 
-This project depends on Python 2.7 or 3.5+, and the Mercurial >= 4.6
-package (>= 5.2, if Python 3.5+). If Python is not installed, install
-it before proceeding. The Mercurial package can be installed with `pip
-install mercurial`.
+This project depends on Python 3.5+, and the Mercurial >= 5.2 package.
+If Python is not installed, install it before proceeding. The Mercurial
+package can be installed with `pip install mercurial`.
 
 On windows the bash that comes with "Git for Windows" is known to work
 well.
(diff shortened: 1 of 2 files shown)


frej:
Thanks for the patch @onlined, but as `importlib.util.find_spec` and friends were introduced in Python 3.4 this patch breaks Python 2.7 compatibility. As far as I can tell, there isn't a simple and elegant way to stay compatible with 2.7, can you?

Otherwise, as I no longer run a machine with 2.7, Github doesn't support it in the CI and 2.7 has been EOL since more than three years, I'm inclined to stop supporting 2.7. Do you want to update the documentation to no longer claim Python 2.7 compatibility in this patch  (and also remove the 2.7 compatibility code we have in the tree in a separate commit)? Or do you want me to do that, and then rebase your commit?

onlined:
I added the documentation update, I think we can merge this as-is and I can remove the compatibility code in a separate PR.

onlined:
I opened #312 for removing the compatibility code. It is still draft, I'll change its status after this PR is merged and I rebase it on top of new `master`

frej:
Thank you for your contribution @onlined. Feel free to un-draft #312, I have reviewed it and will merge it as soon as it is refreshed.

onlined:
Thanks! Un-drafted #312.

PR #312, opened 2023-11-12
Remove Python 2 compatibility code

onlined:
Python 2 support was removed recently, so we don't need the compatibility code anymore.


frej:
Thank you for your contribution @onlined!

PR #320, opened 2024-02-19
Fix escape in regular expression

sth:
The string used to create the regular expression should contain literal backslashes, so they need to be escaped. Python 12 warns about it:

```
hg-fast-export.py:231: SyntaxWarning: invalid escape sequence '\.'
```


frej (approved):
LGTM.

PR #321, opened 2024-02-20
Don't add file if plugin sets content to `None`

sth:
I was looking for a way to filter the files that get included in the new repository. Currently a plugin can *modify* the content, but not completely remove it.

A straight forward way for a plugin to show it doesn't want the file is to set `file_data['data']` to `None`. In the main script this can also easily be handled: Only output the file if its data is not `None`.

The change is quite minimal and shouldn't interfere with any existing code.

### DIFF
--- a/t/file_data_filter-removefiles.t
+++ b/t/file_data_filter-removefiles.t
@@ -1,0 +1,90 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 Felipe Contreras
+# Copyright (c) 2023 Frej Drejhammar
+# Copyright (c) 2024 Stephan Hohe
+#
+# Check that files that file_data_filter sets to None are removed from repository
+#
+
+test_description='Remove files from file_data_filter plugin test'
+
+. "${SHARNESS_TEST_SRCDIR-$(dirname "$0")/sharness}"/sharness.sh || exit 1
+
+check() {
+	echo "$3" > expected &&
+	git -C "$1" show -q --format='%s' "$2" > actual &&
+	test_cmp expected actual
+}
+
+git_create() {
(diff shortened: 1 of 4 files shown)


frej:
I like the idea, but I would like to see a test case where a file is removed using the plugin. The test case should also check that nothing bad happens when a plugin-deleted file is removed in a later mercurial commit.

sth:
Added a test that adds, modifies and removes a file in the hg repository and uses a plugin to remove that file when converting to git. The resulting git repository then doesn't contain any references to the "bad" files.

There are some empty commits in the resulting repository, because all the changes done in those commits were to "bad" files. This has no negative consequences despite maybe looking a bit odd at first sight. But it makes sense since all the changes in those commits were to files that got removed.

t/file_data_filter-removefiles.t
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 Felipe Contreras
+# Copyright (c) 2023 Frej Drejhammar
+# Copyright (c) 2023 Stephan Hohe
```
frej, line 5: Not 2024?

frej (approved):
LGTM apart from the copyright year...

t/file_data_filter-removefiles.t
sth: Yes, you are absolutely right...

frej:
Please squash the fix into the commit for the test case. Unless you protest, I can do that when I get time to start merging later today.

sth:
I squashed it.

frej:
Thank you for your contribution @sth.

PR #333, opened 2024-07-06
Consistently terminate commit messages with LF

thaliaarchi:
When the length logic for fast-import 'data' commands was updated in 4c10270 (Fix data handling, 2023-03-02), one branch was missed, so commit messages now do not have a final LF appended in most cases. This changed the longtime behavior, which had been consistent since the first commit of hg2git, 9832035 (Initial import, 2007-03-06), and is expected by some applications which compare against old conversions from Mercurial[[1]](https://github.com/thaliaarchi/repo-archival/blob/main/scripts/inferno/archive.sh).

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -281,7 +281,7 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
     parents = commit_data['parents']
     author = commit_data['author']
     user = commit_data['committer']
-    desc = commit_data['desc'] + b'\n'
+    desc = commit_data['desc']
 
   if len(parents)==0 and revision != 0:
     wr(b'reset refs/heads/%s' % branch)
@@ -291,7 +291,7 @@ def export_commit(ui,repo,revision,old_marks,max,count,authors,
   if sob:
     wr(b'author %s %d %s' % (author,time,timezone))
   wr(b'committer %s %d %s' % (user,time,timezone))
-  wr_data(desc)
+  wr_data(desc + b'\n')
 
   man=ctx.manifest()
 
--- a/t/file_data_filter-removefiles.expected
+++ b/t/file_data_filter-removefiles.expected
@@ -8,20 +8,23 @@ commit refs/heads/master
 mark :2
 author Grevious Bodily Harmsworth <gbh@example.com> 1679014800 +0000
 committer Grevious Bodily Harmsworth <gbh@example.com> 1679014800 +0000
-data 2
-r0M 100644 :1 good_a.txt
+data 3
+r0
+M 100644 :1 good_a.txt
 
 commit refs/heads/master
 mark :3
 author Grevious Bodily Harmsworth <gbh@example.com> 1679018400 +0000
 committer Grevious Bodily Harmsworth <gbh@example.com> 1679018400 +0000
-data 2
-r1from :2
+data 3
+r1
+from :2
 
 commit refs/heads/master
 mark :4
 author Grevious Bodily Harmsworth <gbh@example.com> 1679022000 +0000
 committer Grevious Bodily Harmsworth <gbh@example.com> 1679022000 +0000
-data 2
-r2from :3
+data 3
+r2
+from :3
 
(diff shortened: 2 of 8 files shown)


thaliaarchi:
This changes the expected output of the tests. Note, though, that `t/` was created after this bug was introduced, so they have never reflected this behavior. I would have updated them myself here, but tests currently fail on master for reasons I don't understand.

frej:
> This changed the longtime behavior, which had been consistent since the first commit of hg2git

Good catch! That should be fixed, no argument there. But we should have the test cases updated in the same commit to stay bisectable.

> I would have updated them myself here, but tests currently fail on master for reasons I don't understand.

I can't reproduce any breakage on master locally. I did #334 to rule out that was just a fluke, but it's not broken in the CI either. If you are switching between branches, have you done `git clean -fxd` to be sure that there's not any old files laying around? Running `make TEST_OPTS='--debug --verbose' -C t` will give you more information. The documentation for the test harness is [here](https://github.com/felipec/sharness#usage).

thaliaarchi:
Thanks for the pointer about `--debug` and `--verbose`. With that, this error is printed, which was hidden before:

> Error: The option core.ignoreCase is set to true in the git
> repository. This will produce empty changesets for renames that just
> change the case of the file name.
> Use --force to skip this check or change the option with
> git config core.ignoreCase false

I have `core.ignoreCase=true` globally configured, because I unfortunately setup my filesystem as case-insensitive long ago. The problem is simply that the tests don't override this. Rather than setting it in the tests' repos as I have done in my own usage of hg-fast-export, I've pushed a patch to not rely on the user's configuration and instead run git fast-import with `ignoreCase` disabled inline. Now users have one less error to worry about. I don't think there's a single use case for wanting path collisions with hg-fast-export, that wouldn't be more correctly solved with a git-filter-repo pass afterwards.

I'm surprised that git fast-import has this issue in the first place, since I'm familiar with the code from contributing patches to it, but, sure enough, it uses variably case-sensitive path comparisons:

```c
int fspathncmp(const char *a, const char *b, size_t count)
{
	return ignore_case ? strncasecmp(a, b, count) : strncmp(a, b, count);
}
```

thaliaarchi:
Should I have left the `core.ignoreCase` change out of this PR? Tests pass for me locally with those changes and should remain working with CI (though I think you need to approve the workflow, for some reason, for the tests to run).

hg-fast-export.sh
```diff
@@ -78,9 +78,9 @@ Options:
 	-o <name> Use <name> as branch namespace to track upstream (eg 'origin')
 	--hg-hash Annotate commits with the hg hash as git notes in the
                   hg namespace.
-	-e <encoding> Assume commit and author strings retrieved from
```
frej, line 81: Please don't include unrelated whitespace changes.

hg-fast-export.sh
```diff
@@ -203,7 +186,7 @@ $(
   } | \
   {
     _e2=0
-    git fast-import $GFI_OPTS --export-marks="$GIT_DIR/$PFX-$SFX_MARKS.tmp" 3>&- || _e2=$?
+    git -c core.ignoreCase=false fast-import $GFI_OPTS --export-marks="$GIT_DIR/$PFX-$SFX_MARKS.tmp" 3>&- || _e2=$?
```
frej, line 189: We have been supporting mangling to a case-insensitive FS since the beginning (although with sometimes surprising results), so I'm not happy about suddenly removing that functionality. I think that we should force `core.ignoreCase=false` in the tests, but not force that on the user.

frej (changes requested):
https://github.com/frej/fast-export/pull/333/commits/f947189dcc1952ceeff04b6d0b8a8f040200740a looks good, but I don't think we should force the user to use `core.ignoreCase=false`. As we leave it to the user to run `git init` they should be free to configure such things without fast-export overriding it.

thaliaarchi:
Thanks for the review. I've reverted the changes to `core.ignoreCase` behavior and updated only tests.

frej (approved):

frej:
Thank you for your contribution @thaliaarchi.

PR #335, opened 2025-03-06
Converted from in commit message

kokostek:
Just a little plugin I've found useful for preserving revision hashes after this (rather unusual) conversion chain:
* take original mercurial repo;
* `hg convert --config convert.hg.saverev=True ...` - maybe wipe some old history, or smth like that, preserving original hashes in `extra`;
* `hg-fast-export.sh ... --plugin converted_from_in_commit` - convert to git, appending original hashes from `extra` to commit messages.


frej:
Thank you for the contribution, but I think it too specialized to be of general interest as it doesn't show of a new aspect of the plugin interface.

PR #340, opened 2025-07-10
Fix TypeError in hg2git.get_branch when using the --origin option

igt-gteubl:
When using the `--origin` option, a `TypeError` is raised:
```
  File "C:\Dev\git-migration\fast-export\hg-fast-export.py", line 728, in <module>
    sys.exit(hg2git(options.repourl,m,options.marksfile,options.mappingfile,
  File "C:\Dev\git-migration\fast-export\hg-fast-export.py", line 547, in hg2git
    if not verify_heads(ui,repo,heads_cache,force,ignore_unnamed_heads,branchesmap):
  File "C:\Dev\git-migration\fast-export\hg-fast-export.py", line 492, in verify_heads
    b=get_branch(b)
  File "C:\Dev\git-migration\fast-export\hg2git.py", line 73, in get_branch
    return origin_name + b'/' + name
TypeError: can only concatenate str (not "bytes") to str

```

This PR fixes this by converting the `origin_name` to `bytes` in `set_origin_name` so that `get_branch` does not crash when trying to concatenate `str` with `bytes`.

### DIFF
--- a/hg2git.py
+++ b/hg2git.py
@@ -23,7 +23,7 @@ user_clean_re=re.compile(b'^["]([^"]+)["]$')
 
 def set_default_branch(name):
   global cfg_master
-  cfg_master = name.encode('utf8') if not isinstance(name, bytes) else name
+  cfg_master = name.encode('utf8')
 
 def set_origin_name(name):
   global origin_name


hg2git.py
```diff
@@ -27,7 +27,7 @@ def set_default_branch(name):
 
 def set_origin_name(name):
   global origin_name
-  origin_name = name
+  origin_name = name.encode('utf-8') if not isinstance(name, bytes) else name
```
frej, line 30: Why the conditional coercion to bytes, `set_origin_name()` is only called from the command line handling so the name will always be a string?

frej (commented):
Good catch!

Please massage the explanation from the PR into a [nice stand alone commit message](https://chris.beams.io/posts/git-commit/)). Also, could I convince you to add a small test case to ensure that this bug stays fixed?

hg2git.py
igt-gteubl: I did the conversion analogous to the one already there in `set_default_branch` directly above. I though it might be a good idea to have both command line options behave in the same way if someone refactors how they are called later on.

I'm relatively new to git (hence the use of your tool), but I'll try to adhere to the `Submitting Patches` sections with my next commit. 

As for the test case, I'll try to whip one up once I get the hang out of how those work.

igt-gteubl: Hope it's ok now.

PR #341, opened 2025-07-25
Fix UnboundLocalError with plugins and largefiles

igt-gteubl:
When Plugins are used in a repository that contains largefiles, the following exception is thrown as soon as the first largefile is converted:

```
  File "fast-export/hg-fast-export.py", line 728, in <module>
    sys.exit(hg2git(options.repourl,m,options.marksfile,options.mappingfile,
  File "fast-export/hg-fast-export.py", line 581, in hg2git
    c=export_commit(ui,repo,rev,old_marks,max,c,authors,branchesmap,
  File "fast-export/hg-fast-export.py", line 366, in export_commit
    export_file_contents(ctx,man,modified,hgtags,fn_encoding,plugins)
  File "fast-export/hg-fast-export.py", line 222, in export_file_contents
    file_data = {'filename':filename,'file_ctx':file_ctx,'data':d}
UnboundLocalError: local variable 'file_ctx' referenced before assignment
```

This commit fixes the error by:

 * initializing the file_ctx before the largefile handling takes place
 * Providing a new `is_largefile` value for plugins so they can detect if largefile handling was applied (and therefore the file_ctx object may no longer be in sync with the git version of the file)


frej (commented):
Looks good, but do I dare asking for a test-case? As largefile support was added before fast-export had a test suite, there isn't much to go on, but ad96531 contains a minimal stand alone test-case. You could copy it and extend it to create and convert largefiles.

README.md
```diff
@@ -188,7 +188,7 @@ values in the dictionary after filters have been run are used to create the git
 commit.
 
 ```
-file_data = {'filename':filename,'file_ctx':file_ctx,'data':file_contents}
+file_data = {'filename':filename,'file_ctx':file_ctx,'data':file_contents, is_largefile:largefile_status}
```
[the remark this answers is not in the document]
igt-gteubl: Test case added and README.md fixed (hopefully).

t/largefile_plugin.t
```diff
@@ -0,0 +1,69 @@
...
+	hg add a.txt &&
+	hg add --large b.txt &&
+	hg commit -d "2023-03-17 01:00Z" -m "r0"
+	)
+}
+
+setup
+
+test_expect_success 'largefile and plugin' '
+#	test_when_finished "rm -rf hgrepo gitrepo" &&
```
frej, line 54: Something left over from testing? The test should clean up its scratch area when done.

igt-gteubl: Dang. Sorry I missed this. Apparently the `sharness` module cleans it up anyway (the working directory stays clean even without this - at least when I run the tests, that's why I missed it), but I'll change it for good measure.

frej: Looks good now. Thank you for your contribution, especially for taking the time to cook up a test case.

PR #351, opened 2026-03-22
Fix: Largefile with CRLF in .hglf file

zingsheim:
Files in .hglf could contain CRLF instead of just LF line endings. The largefile export was confused by this while reading the hash.

Fixed by stripping all whitespaces from the hash.

### DIFF
--- a/hg-fast-export.py
+++ b/hg-fast-export.py
@@ -172,7 +172,7 @@ def largefile_orig_name(filename):
 
 def largefile_data(ctx, file, filename):
   lf_file_ctx=ctx.filectx(file)
-  lf_hash=lf_file_ctx.data().strip(b'\n')
+  lf_hash=lf_file_ctx.data().strip()
   sys.stderr.write("Detected large file hash %s\n" % lf_hash.decode())
   #should detect where the large files are located
   file_with_data = lfutil.findfile(ctx.repo(), lf_hash)


frej:
This isn't the first time HG is sloppy about verifying internal data, sigh!

What about an addition to the test-suite?

zingsheim:
> What about an addition to the test-suite?

In order to reproduce it you probably have to create a hg repository on Windows with a specific setting of the line ending handling changed at the same time as the largefile was added (but this is only a guess since I only saw the failing export). I do not know how to create such a test. Especially, since I do not have a Windows version available myself.

On the repository which was supposed to be converted, the fix did its job.

For ordinary repositories, created and exported on Unix, there is already a test, (see main.t "hg large file")

frej:
> > What about an addition to the test-suite?
> In order to reproduce it you probably have to create a hg repository on Windows with a specific setting of the line ending handling changed at the same time as the largefile was added (but this is only a guess since I only saw the failing export). I do not know how to create such a test. Especially, since I do not have a Windows version available myself.

I understand you pain. I'm not interested in adding Wintendo to the test suite. 

> On the repository which was supposed to be converted, the fix did its job.

That's good enough for me.

> For ordinary repositories, created and exported on Unix, there is already a test, (see main.t "hg large file")

Yes, I remember you previous contribution which included tests, that's why I asked.

I'll merge this as is, thank you for being a returning contributor.

